PR #4, opened 2012-11-06
ADD: interval to table

christianbrodbeck:
Adds the cluster time interval to the as_table method.

### DIFF
--- a/eelbrain/analyze/testnd.py
+++ b/eelbrain/analyze/testnd.py
@@ -509,10 +509,15 @@ class cluster_anova:
                 for i in np.argsort(ps):
                     c = cs[i]
                     p = c.properties['p']
+                    interval = np.nonzero(c.x)[0]
+                    c_start = format(c.time[interval[0]], '1.3f')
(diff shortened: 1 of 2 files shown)


eelbrain/analyze/testnd.py
```diff
@@ -509,10 +509,15 @@ def as_table(self, pmax=1.):
                 for i in np.argsort(ps):
                     c = cs[i]
                     p = c.properties['p']
+                    interval = np.nonzero(c.x)[0]
+                    c_start = format(c.time[interval[0]], '1.3f')
+                    c_stop = format(c.time[interval[-1]], '1.3f')
                     if p > pmax:
                         break
                     table.cell(i)
+                    table.cell((c_start,c_stop))
```
christianbrodbeck, line 518: If you add a column to the table, you'd have to specify that when initializing the table (i.e., `fmtxt.Table('lll')`). Also, I am not sure how the tuple will look - it might be simpler to submit a string directly, e.g., `c_start+' - '+c_stop`

PR #6, opened 2012-11-20
Tb common brain

teonbrooks:
Hey,

I just added a modification to the experiment class to accommodate other common brains as a keyword argument in the initialization.


christianbrodbeck:
added in 4f9a107b900e14d3e92bb6bd5638fc21c8c5a09c

PR #9, opened 2012-11-30
ADD: logger

teonbrooks:
Adds a simple logger (dict) to the experiment class to track changes that
occur during experiment session.


teonbrooks:
This no longer needs to be added. Request withdrawn.

PR #12, opened 2012-12-15
Tb plot exp fiff fix

teonbrooks:
Hey,

The plot modification is to incorporate a dict as a kwarg for the plot.uts.clusters. It seems to be alright.

The experiment modification as the option to not iterate over empty list. e.g. I don't have an experiments class attribute so when it goes to set initial values, it raises an error b/c the iteration is said to be out of range (since the experiment = [ ] ).

In the fiff modification, the kwarg dict retains the tstart and tstop entries, which is problematic when being forwarded to the mne.Epochs class. Since these kwargs are being reassigned to tmin and tmax, I just remove them after assignment.

Let me know if you have any questions.

### DIFF
--- a/eelbrain/load/fiff.py
+++ b/eelbrain/load/fiff.py
@@ -623,8 +623,10 @@ def mne_Epochs(ds, i_start='i_start', raw=None,
     kwargs['name'] = name = name.format(name=ds.name)
     if 'tstart' in kwargs:
         kwargs['tmin'] = kwargs['tstart']
+        del kwargs['tstart']
     if 'tstop' in kwargs:
         kwargs['tmax'] = kwargs['tstop']
+        del kwargs['tstop']
 
     events = mne_events(ds=ds, i_start=i_start)
     # epochs with (event_id == None) does not use columns 1 and 2 of events
--- a/eelbrain/load/fiff.py
+++ b/eelbrain/load/fiff.py
@@ -622,11 +622,9 @@ def mne_Epochs(ds, i_start='i_start', raw=None,
 
     kwargs['name'] = name = name.format(name=ds.name)
     if 'tstart' in kwargs:
-        kwargs['tmin'] = kwargs['tstart']
-        del kwargs['tstart']
+        kwargs['tmin'] = kwargs.pop('tstart')
     if 'tstop' in kwargs:
-        kwargs['tmax'] = kwargs['tstop']
-        del kwargs['tstop']
+        kwargs['tmax'] = kwargs.pop('tstop')
 
     events = mne_events(ds=ds, i_start=i_start)
     # epochs with (event_id == None) does not use columns 1 and 2 of events
--- a/eelbrain/plot/uts.py
+++ b/eelbrain/plot/uts.py
@@ -490,7 +490,7 @@ def _plt_uts(ax, layer, color=None, xdim='time', kwargs={}):
 
 class _ax_clusters:
     def __init__(self, ax, layers, color=None, pmax=0.05, ptrend=0.1,
-                 t = t, xdim='time', title=None):
+                 t = {}, xdim='time', title=None):
         Y = layers[0]
 
         if title:
--- a/eelbrain/plot/uts.py
+++ b/eelbrain/plot/uts.py
@@ -485,9 +485,13 @@ class _ax_clusters:
                 title = title.format(name=Y.name)
             ax.set_title(title)
 
-        if tkwargs:
-            t = Y.properties.get('threshold', None)
-            ax.axhline(t, **tkwargs)
+        t = Y.properties.get('threshold', None)
+        if t:
+            if tkwargs:
+                ax.axhline(t, **tkwargs)
+            else:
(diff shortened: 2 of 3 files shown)


eelbrain/load/fiff.py
```diff
@@ -623,8 +623,10 @@ def mne_Epochs(ds, i_start='i_start', raw=None,
     kwargs['name'] = name = name.format(name=ds.name)
     if 'tstart' in kwargs:
         kwargs['tmin'] = kwargs['tstart']
+        del kwargs['tstart']
```
christianbrodbeck: Thanks for noting this, you could make it a one-liner with:

```
kwargs['tmin'] = kwargs.pop('tstart')
```

but it's fine either way

eelbrain/plot/uts.py
```diff
@@ -348,13 +349,19 @@ def __init__(self, epochs, pmax=0.05, ptrend=0.1, t=True,
             Maximum p-value of clusters to plot as solid.
         ptrend : scalar
             Maximum p-value of clusters to plot as trend.
-        t : bool
+        t : dict
             Plot threshold for forming clusters.
+            Contains threshold plotting properties.
+        linestyle : str
```
christianbrodbeck: These arguments are not in the function anymore - you could just say that the t dictionary is going to be submitted as kwargs to matplitlib's axhline

eelbrain/plot/uts.py
```diff
@@ -337,10 +337,11 @@ def _ax_stat(ax, ct, colors, legend_h={},
 
 
 class clusters(_base.subplot_figure):
-    def __init__(self, epochs, pmax=0.05, ptrend=0.1, t=True,
+    def __init__(self, epochs, pmax=0.05, ptrend=0.1, 
                  title="plot.uts.clusters", figtitle=None, axtitle='{name}',
                  cm=_cm.jet, width=6, height=3, frame=.1, dpi=90,
-                 overlay=False):
+                  overlay=False, t = {'linestyle': 'solid', 'color': 'k'}):
```
christianbrodbeck: Is the indent off by one here?

eelbrain/plot/uts.py
```diff
@@ -386,7 +393,7 @@ def __init__(self, epochs, pmax=0.05, ptrend=0.1, t=True,
 
         width = .85
         height = .95 / Nax
-
+
```
christianbrodbeck: Can you remove those spaces? The code formatter should actually do that, there's a setting for it, "Right trim multi-line string literals"

teonbrooks: I  checked and my right trim multitine is enabled. I can't tell if it has solved that issue.

christianbrodbeck: You can save the file and see if it takes effect (if you have "only apply formatting ti changed lines" enabled, edit the line)

eelbrain/plot/uts.py
```diff
@@ -483,10 +485,9 @@ def __init__(self, ax, layers, color=None, pmax=0.05, ptrend=0.1,
                 title = title.format(name=Y.name)
             ax.set_title(title)
 
-        if t is True:
+        if tkwargs:
             t = Y.properties.get('threshold', None)
-        if t:
-            ax.axhline(t, color='k')
+            ax.axhline(t, **tkwargs)
```
christianbrodbeck: If `Y.properties` does not have `'threshold'`, t is going to be `None`, so before `ax.axhline(t, **tkwargs)` you should make sure t is not None

eelbrain/plot/uts.py
```diff
@@ -348,13 +349,14 @@ def __init__(self, epochs, pmax=0.05, ptrend=0.1, t=True,
             Maximum p-value of clusters to plot as solid.
         ptrend : scalar
             Maximum p-value of clusters to plot as trend.
-        t : bool
+        t : dict
+            Contains kwargs for matplotlib axhline for threshold plotting.
```
christianbrodbeck, line 353: You might add that the threshold can be disabled with `t=False`

christianbrodbeck:
Other than the comments it looks good to me!

eelbrain/plot/uts.py
teonbrooks: Would there be a difference between saying  `if t:` and `if t is not None:`?

eelbrain/plot/uts.py
```diff
@@ -483,10 +485,13 @@ def __init__(self, ax, layers, color=None, pmax=0.05, ptrend=0.1,
                 title = title.format(name=Y.name)
             ax.set_title(title)
 
-        if t is True:
-            t = Y.properties.get('threshold', None)
+        t = Y.properties.get('threshold', None)
         if t:
```
christianbrodbeck: > Would there be a difference between saying if t: and if t is not None:?

Yes, `if t` is like `if bool(t)`, and `bool(0) == False` so in the case that the threshold is zero it would not be plotted

christianbrodbeck:
Merged it, thanks!

PR #13, opened 2012-12-21
Tb fix sensor net

teonbrooks:
Hey,

I noticed a few things when working with the sensor net. Let me know what you think. The motivation is to correct for the number of sensors that the sensor net has. The STIM channel was being added to the net and these changes are devised to solve that from happening.

Best


christianbrodbeck:
When I simple load a raw file and create a sensor net I don't get an error. Can you send me some code that reproduces the error?

I don't think the lower part of your changes (line 651 ff.) actually changes anything (?) Instead of extracting picked channels and then looping through them you loop through picks and extract channels in the loop.

Since you deleted the part that looks for picks in the fiff object, is it possible you added the event channel to your fife file's picks? Then you could simply submit the new picks (i.e., mne.fiff.pick_types(info)) to the sensor net initialization.

teonbrooks:
`Can you send me some code that reproduces the error?`
I will forward you the epochs object. It has the STIM as a channel but I didn't add it.

`is it possible you added the event channel to your fife file's picks?`
I didn't but I am wondering if the update to mne has changed it?

`I don't think the lower part of your changes (line 651 ff.) actually changes anything`
Yeah, you're right. It was from when I was trying to debug. That did nothing. lol

christianbrodbeck:
> I didn't but I am wondering if the update to mne has changed it?

I tried from a simple raw file (one form my experiment as well as the sample audvis raw file) with the latest mne version, using

```
raw = mne.fiff.Raw(fname)
s = load.fiff.sensor_net(raw)
print s
```

teonbrooks:
I can't get the file synced in my dropbox. Here's my output though. It
doesn't show up in the raw file but in the epochs, it does.

In [1]: e = process.NMG()

In [2]: ds = e.load_events('R0095')
subject: R0095:
    Read a total of 1 projection items:
        axial-0--0.200-0.600-PCA-01 (1 x 157)  idle
edf: R0095:  97% remains

In [3]: ds = ds[ds['target'] == 'target'][:2]

In [4]: ds = load.fiff.add_mne_epochs(ds, reject = {'mag': 3e-12}, tstart =
-.1, tstop = .6)
Created an SSP operator (subspace dimension = 1)
1 projection items activated
2 matching events found

In [5]: load.fiff.sensor_net(ds['epochs'])
Out[5]: sensor_net([<n=158>], name='fiff-sensors')

In [6]: load.fiff.sensor_net(ds.info['raw'])
Out[6]: sensor_net([<n=157>], name='fiff-sensors')

## 

Teon

christianbrodbeck:
Thanks! I can reproduce it now. The epochs.picks seem to be the picks used to create the epochs, so they do not apply to the channels that are present in the epochs but to the channels that were present in its source.

Now how to fix it - logically it seems to me the sensor_net should reflect the channels that the user gives it, so the user should exclude the event channel before initializing the sensor_net (because the same has to be done to any data used with it). What is the actual use case for initializing a sensor_net from the Epochs? Are you converting an Epochs to an ndvar after loading? In that case it might be worth writing a proper epochs_ndvar function - what do you think?

teonbrooks:
Hey,

In experiment.make_proj_for_epochs on line 652, the sensor net is being initiated with
the epochs object, not the raw fif. I guess for that, it could initiated
with the 'raw' instead of 'epochs'.

Would the epochs_ndvar essentially do what the add_epochs function is
currently doing after the epochs have been extracted?

## 

Teon

christianbrodbeck:
Ok, I added a fix to mne_experiment.make_proj_for_epochs(). Let me know if the problem occurs in other areas. Also, since they obviously modified the proj class, are the projs still applied properly?

> Would the epochs_ndvar essentially do what the add_epochs function is
> currently doing after the epochs have been extracted?

What I had in mind was an analog to stc_ndvar, bit that does not seem relevant for projs.

PR #14, opened 2013-02-18
Tb ui select proj fix

teonbrooks:


eelbrain/vessels/experiment.py
```diff
@@ -1295,6 +1294,7 @@ def ui_select_projs(self, projs, fif_obj, save=True, save_plot=True):
 
             p.close()
             projs = [projs[i] for i in rm]
+            proj_file = self.get('proj-file')
```
christianbrodbeck, line 1297: proj_file is needed to save the projections even if save_plot = False

eelbrain/vessels/experiment.py
```diff
@@ -1274,12 +1274,11 @@ def ui_select_projs(self, projs, fif_obj, save=True, save_plot=True):
...
         for p in projs:
             d = p['data']['data'][0]
-            name = p['desc'][-5:]
+            name = p['desc'][-6:]
             v = ndvar(d, (sensor,), name=name)
             PCA.append(v)
 
-        proj_file = self.get('proj-file')
-        p = plot.topo.topomap(PCA, size=1, title=proj_file)
+        p = plot.topo.topomap(PCA, size=1, title=name)
```
christianbrodbeck, line 1281: Is proj_file as title too long? We could use the basename then. name is only a specific component name

eelbrain/vessels/experiment.py
teonbrooks, line 1297: I was doing this for cases where `save_plot = True` and `save = False`. I guess I have a more general concern. Is there a standard template? I noticed that `proj` is no longer default but `proj-file` is and `rawfif` isn't but `raw-file`.

eelbrain/vessels/experiment.py
teonbrooks, line 1281: I change it based on the above comment. I had not defined `proj-file` in my template so I needed to give it a title based on something else.

eelbrain/vessels/experiment.py
christianbrodbeck, line 1297: yes I've started trying to standardize a little, such that all files would always have the "-file" flag. The reason is that for each file you might also want to specify an option, e.g. which projections file to use.  .. I didn't realize it would affect you since I was thinking you had your own templates :) Maybe we should go over the templates once and settle it? Are you planning on going to the office today?

teonbrooks, line 1297: I think I will. I'm already up now so I guess I should do something with my life. :) I have been going through and have been trying to prune methods from my subclass that exists in the template. I figured that it would be more beneficial and keep things current.

PR #16, opened 2013-09-30
Tb colorspaces fix

teonbrooks:
These are fixes we worked on at my desk.


teonbrooks:
somehow these other commits were added into this request.

PR #21, opened 2015-05-06
beginning to add functions that can get average activity over time, or per ms

LauraGwilliams:
I've made the function save the ds's that are created for each subject to disk, and then load them up as a separate function to make things go quicker. will need to combine them into a single function at some point.


christianbrodbeck:
Feel free to reopen if you're still working on this

PR #23, opened 2015-09-09
add frequency tests to e.make_report

LauraGwilliams:


eelbrain/experiment/_mne_experiment.py
```diff
@@ -1774,38 +1778,60 @@ def load_evoked_freq(self, subject=None, sns_baseline=True,
...
-            inv = self.load_inv(ds_epochs['epochs'])
-            subjects_dir = self.get('mri-sdir')
-            ds = source_induced_power('epochs', model, ds_epochs, src, label,
-                                      None, inv, subjects_dir, frequencies,
-                                      n_cycles=3,
-                                      **self._params['apply_inv_kw'])
-            ds['subject'] = Factor([subject], repeat=ds.n_cases, random=True)
-        return ds
+            ds_stc = self.load_evoked_stc(subject=subject, sns_baseline=sns_baseline,
+                                          morph_ndvar=True)
```
christianbrodbeck: Do you intentionally load averaged data here? The original function was supposed to return average power over epochs, not power in the averages.

eelbrain/experiment/_mne_experiment.py
```diff
@@ -1764,8 +1767,9 @@ def load_evoked(self, subject=None, baseline=None, ndvar=True, cat=None,
 
         return ds
 
-    def load_evoked_freq(self, subject=None, sns_baseline=True,
-                         label=None, frequencies='4:40', **kwargs):
+    def load_evoked_freq(self, subject=None, freq_name=None, sns_baseline=True,
+                         parc=None, labels=None,
+                         hemi=None, source_average=True, **kwargs):
```
christianbrodbeck: I would not add the the parc/labels/hemi/source_average arguments here. parc is a parameter in the experiment class, so you can use `parc=e.get('parc')` below (although any etc loading function would already set the current parc so you should not have to deal with it at all); and subsetting at this point is only really necessary if it's a performance issue, in which case we can deal with it later, otherwise subsetting is easily done outside this function.

eelbrain/experiment/_mne_experiment.py
LauraGwilliams: Hm, the idea was to create one function that loads evoked_freq and another that loads epochs_freq (mirroring load_evoked_stc and load_epochs_stc). Do you think it would be better to have them in the same function?

christianbrodbeck: Hmm you are right we should be careful with terminology… What about these:

load_epochs_stf() for single trial data (stf = source time frequency)
load_evoked_stf() for frequency transformed evoked data (what you added) 
load_induced_stf() for the average of single trial stf

LauraGwilliams: OK sure, I can do that :)

I have a question… I’m trying to make it so that the name of the frequency band we are after can also be retrieved from the experiment definition file. But it seems I have to add something to the eelbrain scripts to enable e.set(frequencies=‘freq_name’). Do you know what I would need to modify to allow for this? (We can talk about this later if you prefer) :)

christianbrodbeck: Yes, you could look at how .parcs works. you would need to 
- add an empty .frequencies dict in which users can define their own frequencies
- a ._frequencies attribute in which you can store pre-defined frequencies
- in **init**, you’d need to add a section which combines them, checks the values and stores them by overwriting the ._frequencies attribute. 
- later, register the ‘frequency’ field

you could search for “parcs” and “parc” to see how these things are done for parcs.

eelbrain/experiment/_mne_experiment.py
```diff
@@ -1800,8 +1824,72 @@ def load_evoked(self, subject=None, baseline=False, ndvar=True, cat=None,
...
+        source_average : bool
+            Whether to average over sources in the given ROI. If False,
+            maintains the spatial dimension.
+        mask : bool
+            Discard data that is labelled 'unknown' by the parcellation (only
+            applies to NDVars, default True).
+        """
+
+        # turn off filter settings because we're dealing with frequency data here
+        self.set(raw='clm')
```
christianbrodbeck: This should not be here anymore (we have data_raw now)

eelbrain/experiment/_mne_experiment.py
```diff
@@ -363,6 +366,13 @@ class MneExperiment(FileTree):
                            'labels': {'occipitotemporal': "occipital + temporal"}}}
     parcs = {}
 
+    # Frequencies
+    _freqs = {'gamma': {'wavelet': 'mortlet',
```
christianbrodbeck: Since we currently don't have other options I would not include the wavelet entry for now

eelbrain/experiment/_mne_experiment.py
```diff
@@ -363,6 +366,13 @@ class MneExperiment(FileTree):
...
+                        'frequency_range': np.arange(25,50,2), # lowbound, highbound, step
```
christianbrodbeck: I would say "frequencies" doesn't "range" imply that you have just start and stop values?

eelbrain/experiment/_mne_experiment.py
```diff
@@ -617,6 +627,19 @@ def _rej_epochs(epoch):
...
 
         ########################################################################
+        # frequency
+        user_freqs = self.freqs
+
+        freqs = {}
+        for name, f in chain(self._freqs.iteritems(), user_freqs.iteritems()):
+            if name in freqs:
+                raise ValueError("Frequency %s defined twice" % name)
+
```
christianbrodbeck, line 634: you could do some more checking that all required entries are there

christianbrodbeck:
I added two tiny comments, and if you're busy I can fix them myself, just wanted to sensitivize you to the issues :)

LauraGwilliams:
OK, fixed!! :)

christianbrodbeck:
Thanks!

PR #24, opened 2016-06-09
ENH temporal tests: correction across ROIs

reddigari:


eelbrain/_stats/testnd.py
```diff
@@ -2862,6 +2862,50 @@ def info_list(self, title="Computation Info"):
...
+            self.samples = cdists[0][0].samples
+            dist = {}
+            for i, effect in enumerate(self.effects):
+                if np.any([d[i].n_clusters for d in cdists]):
+                    dist[effect] = np.column_stack([d[i].dist for d in cdists if d[i].dist is not None])
+            if len(dist):
+                dist = {c: d.max(1) for c, d in dist.iteritems()}
+        else:
+            self.samples = cdists[0].samples
+            if np.any([d.n_clusters for d in cdists]):
```
christianbrodbeck: this turns `[d.n_clusters for d in cdists]` into a NumPy array first, slightly more efficient would be `any(d.n_clusters for d in cdists)`

christianbrodbeck: Just to show:

``` py
In [2]: timeit np.any([i for i in xrange(10)])
The slowest run took 11.67 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 5.39 µs per loop

In [3]: timeit any(i for i in xrange(10))
The slowest run took 7.29 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 818 ns per loop
```

eelbrain/_stats/testnd.py
```diff
@@ -2862,6 +2862,50 @@ def info_list(self, title="Computation Info"):
         l.add_item("Permutation time:  %s" % timedelta(seconds=round(self.dt_perm)))
         return l
```
christianbrodbeck, line 2882: pep8: two empty lines before class

christianbrodbeck:
Looks good, although I haven't tried it yet. Do you want to add some tests? I assume you want to keep it private for now?

eelbrain/_stats/testnd.py
```diff
@@ -2862,6 +2862,50 @@ def info_list(self, title="Computation Info"):
...
+            for cl in clusters.itercases():
+                n_larger = np.sum(self.dist[cl['effect']] > np.abs(cl['v']))
+                cluster_p_corr.append(float(n_larger) / self.samples)
+        else:
+            keys.append('p_parc')
+            vs = np.array(clusters['v'])
+            n_larger = np.sum(self.dist > np.abs(vs[:, None]), 1)
+            cluster_p_corr = n_larger / self.samples
+        clusters['p_parc'] = Var(cluster_p_corr)
+        clusters = clusters[keys]
```
christianbrodbeck, line 2924: I see, it would probably be useful to have a `Dataset.insert()` method to insert an item at a location other than last?

reddigari:
Potential issue: if an effect has no clusters to test in a region, no permutations are done. Then, if that effect has clusters in another region, the merged permutation distribution only has the permutations from the latter region, and the p-value remains uncorrected. Is that right?

christianbrodbeck:
Right, good catch.

So we need a setting to force the test to do the permutations even if no cluster is found. What about adding a `force_permutation` boolean parameter to the `testnd` tests?

Do you know how to rebase? I've set up Travis so once you rebase on the latest master you should get tests run automatically whenever you push a branch.

reddigari:
I know how to rebase, but I just merged your latest changes into my fork. What should I have done instead?

christianbrodbeck:
Usually if you want to get changes from the master into your PR the procedure is to rebase the PR on the latest master, so that the history is a straight branch coming out of the master. I am not sure how the pull request will look after you merge into it changes from the master, but I think it can get messy. You could just reset your branch to the commit right before the merged and then rebase.

reddigari:
I've been working on the master rather than a separate branch. Is it still possible to rebase my changes onto your master? (Sorry for being so git-naive.)

reddigari:
Ah, I guess `git pull --rebase upstream master` is what I need?

christianbrodbeck:
Yes that should work if you haven't already merged changes from upstream into your master (technically you've been working on a different branch, i.e. reddigari:master is a different branch from my christianbrodbeck:master). In case you don't have a tool to visualize the tree I recommend https://rowanj.github.io/gitx/. Usually you don't _need_ to rebase a PR unless there are conflicts with changes that happened in the master in the meantime.

reddigari:
Thanks. I reset to before the merge, and pulled to rebase. All good.

I will work on adding a `force_permutation` option to the `testnd` tests.

How should I go about adding tests to check that this works?

christianbrodbeck:
Tests go into the `tests/test_x.py` files located in the same subfolder as the relevant functionality. Tests for this would probably go into `eelbrain/_stats/tests/test_testnd.py`. For example, see https://github.com/christianbrodbeck/Eelbrain/blob/master/eelbrain/_stats/tests/test_testnd.py#L287. When the tests are run, each `test_x()` function gets executed and the test reports if any of the assert statements failed. So you would probably add a function to this file called `test_merged_cluster_dist` in which you run some permutation cluster tests, merge their dists and then check that the results look like they should. Probably you would use `datasets.get_uts()` to create data. the data returned by that function should always the identical, so you can hard code when you expect specific results.

christianbrodbeck:
And that would be how you run only specific tests:

``` bash
ISRWKAVW2269D ~/Code/Eelbrain$ cd eelbrain/_stats/tests
ISRWKAVW2269D ~/Code/Eelbrain/eelbrain/_stats/tests$ nosetests test_testnd
..S........
----------------------------------------------------------------------
Ran 11 tests in 2.375s

OK (SKIP=1)
```

reddigari:
Eliminating the extra step as you suggested when checking for clusters doesn't seem to evaluate to True/False:

```
ipdb> cdists
[[<ClusterDist: 1 clusters (0:00:00), 10000 permutations (0:00:18)>, <ClusterDist: 0 clusters (0:00:00)>, <ClusterDist: 0 clusters (0:00:00)>, <ClusterDist: 2 clusters (0:00:00), 10000 permutations (0:00:18)>, <ClusterDist: 0 clusters (0:00:00)>, <ClusterDist: 0 clusters (0:00:00)>, <ClusterDist: 0 clusters (0:00:00)>], [<ClusterDist: 0 clusters (0:00:00)>, <ClusterDist: 0 clusters (0:00:00)>, <ClusterDist: 0 clusters (0:00:00)>, <ClusterDist: 0 clusters (0:00:00)>, <ClusterDist: 0 clusters (0:00:00)>, <ClusterDist: 0 clusters (0:00:00)>, <ClusterDist: 1 clusters (0:00:00), 10000 permutations (0:00:18)>]]
ipdb> np.any(d[i].n_clusters for d in cdists)
<generator object <genexpr> at 0x1134694b0>
ipdb> np.any([d[i].n_clusters for d in cdists])
False
ipdb> i
1
ipdb> np.any(d.n_clusters for d in cdists[0])
<generator object <genexpr> at 0x113469730>
ipdb> np.any([d.n_clusters for d in cdists[0]])
True
```

christianbrodbeck:
use `any` instead of `np.any`

reddigari:
I've begun by just altering `testnd.anova` to force permutations. If `force_permutation` is `True`, it sets the `do_permutation` parameter to `True` and calls `_create_dist` on each effect's cluster distribution. But when the test runs, the values in the distributions are all zero (which I assume means no permutations were actually run). Can you point me in the right direction or tell me what I'm missing about `run_permutations_me`?

christianbrodbeck:
I would 
- Add a `force_permutation` argument to the _ClusterDist class (default `False`), and store the argument somewhere around here   https://github.com/christianbrodbeck/Eelbrain/blob/master/eelbrain/_stats/testnd.py#L2129 to `self.force_permutation = force_permutation`. 
- Then change https://github.com/christianbrodbeck/Eelbrain/blob/master/eelbrain/_stats/testnd.py#L2197 to `if self.force_permutation or self.samples and n_clusters:`

I think that should already be enough to force the permutation, the test function would just have to forward the force_permutation argument, there might then be errors later on in cases which having no cluster but still did permutations...

Does that help?

That should get it working, if it does we will also have to check that saving and loading still works across versions.

christianbrodbeck:
Also feel free to push temporary version of your code if you have questions, you can always later fix it up and update the PR with `git push --force`

reddigari:
That helps, thanks!

reddigari:
I think it's almost ready. The tests check that `_MergedTemporalClusterDist` works with anovas, the three t-tests, and t-contrasts. It also ensures an error is raised if `force_permutation` is used with 0 samples.

How do we test whether saving/loading results will still work between versions?

eelbrain/_stats/testnd.py
```diff
@@ -2194,7 +2206,9 @@ def add_original(self, stat_map):
         self.has_original = True
         self.dt_original = self._t0 - self._init_time
         self._original_param_map = stat_map
-        if self.samples and n_clusters:
+        if self.force_permutation and not self.samples:
+            raise ValueError("Cannot force permutations unless samples is non-zero.")
```
christianbrodbeck: I would make that check in `__init__`, the earlier you can raise an error for invalid arguments the better

eelbrain/_stats/tests/test_testnd.py
```diff
@@ -524,3 +524,57 @@ def test_cwt():
     cluster = res.clusters.sub("p == 0")
     assert_array_equal(cluster['frequency_min'], 10)
     assert_array_equal(cluster['frequency_max'], 19)
+
+def test_merged_temporal_cluster_dist():
+
```
christianbrodbeck: Can you add a short docstring describing the test? it's what 's displayed when running the tests

eelbrain/_experiment/mne_experiment.py
```diff
@@ -4444,11 +4445,24 @@ def make_report_rois(self, test, parc=None, pmin=None, tstart=0.15, tstop=None,
         model = self._tests[test]['model']
         colors = plot.colors_for_categorial(ds.eval(model))
         test_kwargs = self._test_kwargs(samples, pmin, tstart, tstop, ('time',), None)
+        label_results = {}
+        force_permutation = (test_kwargs['samples'] and len(labels_lh + labels_rh) > 1)
```
christianbrodbeck: Actually per doc string `samples` has to be `> 0` so you could just raise an error if it is not. I assume other things might break down below if it is 0.

christianbrodbeck: also `len(labels_lh) + len(labels_rh)` would be more efficient because it does not rely on initializing a new potentially long list

eelbrain/_experiment/mne_experiment.py
```diff
@@ -4444,11 +4445,24 @@ def make_report_rois(self, test, parc=None, pmin=None, tstart=0.15, tstop=None,
...
+        for label in labels_lh + labels_rh:
+            res = self._make_test(ds[label_keys[label]], ds, test, test_kwargs,
+                                  force_permutation)
+            label_results[label_keys[label]] = res
```
christianbrodbeck: You could just use label names as keys here (i.e., `label_results[label] = res`). The reason I generate those label_keys is because label names are not necessarily valid `Dataset` keys (containing dots and such) and the data is stored in the dataset `ds`.

christianbrodbeck:
Other than these minor comments this looks good to me.

> How do we test whether saving/loading results will still work between versions?

You mean IO for cluster test results or for `_MergedTemporalClusterDist`?

reddigari:
IO for the results.

> we will also have to check that saving and loading still works across versions.

I thought you meant adding `force_permutation` to the result objects might have consequences for previously saved tests. No?

christianbrodbeck:
It might have, but in the current implementation `force_permutation` is not used in the `Result`s anywhere that would be accessed after running the test (and they can only be saved after running it) so that we don't need to worry about.

Your changes look good, let me know if you want it merged or if you are still working on it.

You could add a note to the Changes for the next release at https://github.com/christianbrodbeck/Eelbrain/blob/master/doc/source/changes.rst if you want

reddigari:
The last thing I'm going to do is change significance indicators to reflect the corrected p-value. That's the behavior for spatiotemporal results, right?

christianbrodbeck:
spatiotemporal reports show the uncorrected p value in the title (to alert you to anything interesting in the table of contents) but list the corrected value right underneath. 

You’re right that it should probably be the same for both temporal and spatiotemporal tests, but I don’t feel strongly which way it is so if you prefer it the other way feel free to change it.

reddigari:
I'll leave it as is, then. Can you take a look at the Travis build failure? It seems to have failed without doing anything.

In the changes file, should I start a new section for 0.24?

christianbrodbeck:
> I'll leave it as is, then. Can you take a look at the Travis build failure? It seems to have failed without doing anything.

It's been happening occasionally, I am not sure why, they did say that they were having some issues with Mac builds

> In the changes file, should I start a new section for 0.24?

yes

reddigari:
Ready to be merged if you're okay with it. Thanks for all the help!

christianbrodbeck:
Thanks for the contribution! Do you want to add your name after the line describing the change?

reddigari:
Nope

christianbrodbeck:
Thanks!

PR #25, opened 2016-11-07
FIX MneExperiment.load_evoked_stc(): check that mask is boolean

reddigari:
Prevents one from passing the parcellation name as the mask and becoming confused at the results.


christianbrodbeck:
I've merged this into the r/0.24. In my master you can already supply a string (parc name). I've pushed my master now, but I did some major refactoring so if you use it beware that there might be new bugs, but you're welcome to test it.

PR #27, opened 2018-07-02
allow parsing 'vol-10' correctly

proloyd:
It could not read the grade correctly, because from 'vol-10' src it reads kind = 'vol' and grade='1' instead of '10'. Fixed this.


christianbrodbeck:
This line is obsolete in the branch with `VolumeSourceSpace`, I just have not pushed these changes into `master` yet... in the `proloy27` branch you should not encounter this error, no?

PR #28, opened 2018-08-28
Adding glassbrain feature

proloyd:
`GlassBrain` Class to hold 2d projections of an ROI/mask image (by default 3 projections: Frontal, Axial, and Lateral).
`GlassBrain` subclass comes with Eelbrain GUI integration and methods to visualize data in :class:`NDVar` format.

Source: nilearn.plotting.glassbrain (http://nilearn.github.io/modules/generated/nilearn.plotting.plot_glass_brain.html)


eelbrain/_data_obj.py
```diff
@@ -9199,6 +9199,7 @@ def _dim_index(self, index):
             return SourceSpaceBase._dim_index(self, index)
 
 
+
```
christianbrodbeck: You don't need this line :)

eelbrain/plot/_glassbrain.py
```diff
@@ -0,0 +1,437 @@
...
+
+
+def _get_mne_source_space(ndvar):
+    """
+    load a mne source space from a ndvar source space
+    """
+    # mne src
+    from mne import read_source_spaces
+    return read_source_spaces (join(ndvar.source.subjects_dir, ndvar.source.subject, 'bem',
+                                         '%s-%s-src.fif' % (ndvar.source.subject, ndvar.source.src)))
```
christianbrodbeck: You can use `ndvar.source.get_source_space()` for this

eelbrain/plot/_nifti_utils.py
```diff
@@ -0,0 +1,112 @@
...
+    if mri_resolution:
+        affine = src[0]['vox_mri_t']['trans'].copy()
+    else:
+        affine = src[0]['src_mri_t']['trans'].copy()
+    if dest == 'mri':
+        affine = np.dot(src[0]['mri_ras_t']['trans'], affine)
+    affine[:3] *= 1e3
+
+    try:
+        import nibabel as nib  # lazy import to avoid dependency
```
christianbrodbeck: No need, it's already a hard dependency...

eelbrain/plot/_glassbrain.py
```diff
@@ -0,0 +1,430 @@
...
+
+    Parameters
+    ----------
+    ndvar : NDVar  ([case,] time, source[, space])
+        Data to plot; if ``ndvar`` has a case dimension, the mean is plotted.
+        if ``ndvar`` has a space dimension, the norm is plotted.
+    dest : 'mri' | 'surf'
+        If 'mri' the volume is defined in the coordinate system of
+        the original T1 image. If 'surf' the coordinate system
+        of the FreeSurfer surface is used (Surface RAS).
```
christianbrodbeck, line 57: How does this affect the figure? Is it just about the axes coordinates?

eelbrain/plot/_nifti_utils.py
```diff
@@ -0,0 +1,114 @@
...
+                         % src_type)
+
+    if ndvar.has_dim('space'):
+        ndvar = ndvar.norm('space')
+
+    if ndvar.has_dim('time'):
+        data = ndvar.get_data(('source', 'time'))
+    else:
+        data = ndvar.get_data('source')
+        if data.ndim == 1:
```
christianbrodbeck: This will always be true. You could add the dimension in the `get_data()` call with `ndvar.get_data(('source', np.newaxis))`.

eelbrain/plot/_glassbrain.py
```diff
@@ -0,0 +1,430 @@
...
+    mri_resolution: bool
+        It True the image is saved in MRI resolution.
+        WARNING: if you have many time points the file produced can be
+        huge.
+    black_bg : boolean, optional
```
christianbrodbeck: Calling it "file" might be misleading since as far as I saw there is no file being generated (?), could you just say that memory consumption might be high and drawing slow?

christianbrodbeck (commented):
Just a few initial comments, I'll try it out next.

The tests failed because of a couple of style issues, you can see the specific issues by scrolling to the bottom. Let me know if they're not clear enough.

christianbrodbeck:
Just noticed that you've been working on it already :)

eelbrain/plot/_glassbrain.py
proloyd: This has something to do do with the axes coordinates, I copied that thing from nilearn.

eelbrain/_data_obj.py
proloyd: fixed now

eelbrain/plot/_glassbrain.py
proloyd: Okay, I was not aware of this. I fixed it now.

eelbrain/plot/_nifti_utils.py
proloyd: Sure thing. Removed the lazy import now.

eelbrain/plot/_glassbrain.py
proloyd: Actually, the utility function was primarily for saving the image in nifty format, so I kept the original documentation. Now I changed it to memory consumption.

proloyd:
Hi Christan,
As you can see the windows and linux build were completed successfully. But the MacOSX build is getting halted at 'eelbrain/_wxgui/tests/test_select_components.py' . I am not sure why it is so. Any clue?

christianbrodbeck:
Could it be due to `nilearn` importing `pyplot` (see [here](https://github.com/nilearn/nilearn/search?q=pyplot&unscoped_q=pyplot))? `pyplot` does some GUI initialization and should only be imported in a lazy fashion.

christianbrodbeck:
You should not need to modify the CI scripts, did you try just making the `nilearn` imports which also import `pyplot` lazy?

proloyd:
Actually, I was not very comfortable with the things! Now I reverted back all of the Cl scripts to the original one. Setting up the backend for matplotlib did the trick for macos. But unfortunately, building in win32 failed because of test_anova. I will be running the build in win32 once again to see if I can reproduce that.

proloyd:
So, fixing matplotlib backend in osx resolves the issue.

proloyd:
Okay, I merged your request for lazy import. Thanks for helping me out. I am totally clueless about this gui stuff. :)

eelbrain/plot/_glassbrain.py
christianbrodbeck: If it included copied code I think we should make sure it’s acknowledged with the license, like https://github.com/christianbrodbeck/Eelbrain/blob/master/eelbrain/mne_fixes/__init__.py

eelbrain/plot/_glassbrain.py
```diff
@@ -0,0 +1,423 @@
...
+    outline = cbar.outline.get_xy()
+    outline[:2, 1] += cbar.norm(cbar_vmin)
+    outline[2:6, 1] -= (1. - cbar.norm(cbar_vmax))
+    outline[6:, 1] += cbar.norm(cbar_vmin)
+    cbar.outline.set_xy(outline)
+    cbar.set_ticks(new_tick_locs, update_ticks=True)
+
+
+class GlassBrain(TimeSlicer, EelFigure):
+    """`GlassBrain` Class to hold 2d projections of an ROI/mask image
```
christianbrodbeck: "`GlassBrain` Class" is redundant, I would just use something like "Plot 2d projections of a volume source space :class:`NDVar`"

By the way, if you want code style, use "``", not "`" (see [here](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html)).

proloyd:
by the way, nilearn's license.txt also includes licenses for its dependencies, i.e. https://github.com/nilearn/nilearn/blob/master/LICENSE. Shall I include them also?

eelbrain/plot/_glassbrain.py
```diff
@@ -0,0 +1,423 @@
...
+
+    `GlassBrain` subclass comes with Eelbrain GUI integration and
+    methods to visualize data in :class:`NDVar` format.
+    (by default 3 projections: Frontal, Axial, and Lateral).
+
+    Parameters
+    ----------
+    ndvar : NDVar  ([case,] time, source[, space])
+        Data to plot; if ``ndvar`` has a case dimension, the mean is plotted.
+        if ``ndvar`` has a space dimension, the norm is plotted.
```
christianbrodbeck, line 53: Maybe we should not automatically pot the norm, because in the future we might want to add arrows?

eelbrain/plot/_glassbrain.py
christianbrodbeck: "It" -> "If", and saved does not seem too apply, maybe "resampled"?

eelbrain/plot/_glassbrain.py
```diff
@@ -0,0 +1,423 @@
...
+        If 'mri' the volume is defined in the coordinate system of
+        the original T1 image. If 'surf' the coordinate system
+        of the FreeSurfer surface is used (Surface RAS).
+    mri_resolution: bool
+        It True the image is saved in MRI resolution.
+        WARNING: if you have many time points the RAM consumption can be
+        huge.
+    black_bg : boolean, optional
+        If True, the background of the image is set to be black.
+    display_mode : string, optional. Default is 'ortho'
```
christianbrodbeck: The default should be in the description, not the type field

eelbrain/plot/_glassbrain.py
```diff
@@ -0,0 +1,423 @@
...
+    vmin : float
+        Lower bound for plotting, passed to matplotlib.pyplot.imshow
+    vmax : float
+        Upper bound for plotting, passed to matplotlib.pyplot.imshow
+    plot_abs : boolean, optional
+        If set to True (default) maximum intensity projection of the
+        absolute value will be used (rendering positive and negative
+        values in the same manner). If set to false the sign of the
+        maximum intensity will be represented with different colors.
+        See http://nilearn.github.io/auto_examples/01_plotting/plot_demo_glass_brain_extensive.html
```
christianbrodbeck: Can we make it a hyperlink?

eelbrain/plot/_glassbrain.py
christianbrodbeck: "optional" is redundant

eelbrain/plot/_glassbrain.py
```diff
@@ -0,0 +1,423 @@
...
+        "nearest" is faster but can be noisier in some cases.
+    h : scalar
+        Plot height (inches).
+    w : scalar
+        Plot width (inches).
+
+    Notes
+    -----
+    The plotted image should be in MNI space for this function to work
+    properly.
```
christianbrodbeck: So should we actually check that the subject is `fsaverage` and raise an appropriate error if it is not?

eelbrain/plot/_glassbrain.py
```diff
@@ -0,0 +1,423 @@
...
+
+        if ndvar:
+            if ndvar.has_case:
+                ndvar = ndvar.mean('case')
+            # instead check ndvar
+            if ndvar.has_dim('space'):
+                ndvar = ndvar.norm('space')
+
+            self._ndvar = ndvar
+            self.src = ndvar.source.get_source_space()
```
christianbrodbeck: I would make this private as well, i.e. `self._src`, so that we don't crowd the namespace accessible by users

christianbrodbeck (commented):
Hey, I started going through again and noticed I have a lot of small requests to make things more consistent with the rest of Eelbrain, do you prefer me to spell them all out or do you prefer me to just fix them myself? :)

christianbrodbeck:
> by the way, nilearn's license.txt also includes licenses for its dependencies, i.e. https://github.com/nilearn/nilearn/blob/master/LICENSE. Shall I include them also?

That's not necessary since we are not copying from any of those dependencies

proloyd:
Hi Christian, please feel free to fix them, specially the doc-strings. I am kinda newbie in these matter. I will learn them better by looking at your style. :)

eelbrain/plot/_glassbrain.py
proloyd: As far as I understand, MNI is a coordinate system. Does this have something to do with fsaverage?

christianbrodbeck:
Fsaverage is in MNI coordinates

eelbrain/plot/_glassbrain.py
proloyd: Thnaks for sharing the style doc. That is really helpful!

proloyd: I took care of this.

eelbrain/plot/_glassbrain.py
proloyd: But, even then, the activation intensity will be plotted as norm. So, I think we shall keep it this way.

eelbrain/plot/_glassbrain.py
proloyd: fixed to this:
display_mode : Default is 'ortho'
        Choose the direction of the cuts: 'x' - sagittal, 'y' - coronal,
        'z' - axial, 'l' - sagittal left hemisphere only,
        'r' - sagittal right hemisphere only, 'ortho' - three cuts are
        performed in orthogonal directions. Possible values are: 'ortho',
        'x', 'y', 'z', 'xz', 'yx', 'yz', 'l', 'r', 'lr', 'lzr', 'lyr',
        'lzry', 'lyrz'. too. 

Or should list the choices in the description? like 'rh' | 'lh' ?

eelbrain/plot/_glassbrain.py
proloyd: Did that

eelbrain/plot/_glassbrain.py
proloyd: Sure thing.

eelbrain/plot/_glassbrain.py
christianbrodbeck: I mean [like this](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links)

eelbrain/plot/_glassbrain.py
christianbrodbeck: It should be something like:
```
display_mode : str
    Direction of the cuts: 'x' - sagittal, 'y' - coronal,
    'z' - axial, 'l' - sagittal left hemisphere only,
    'r' - sagittal right hemisphere only, 'ortho' - three cuts are
    performed in orthogonal directions (default). 
    Possible values are: 'ortho',
    'x', 'y', 'z', 'xz', 'yx', 'yz', 'l', 'r', 'lr', 'lzr', 'lyr',
    'lzry', 'lyrz'.

```
You could list all choices, but it might be too many

christianbrodbeck (commented):
More generally, we should try to avoid importing anything that starts with an `_` from `nilearn`, those are private functions, and they can be changed without notice, so importing them might likely lead to things breaking in the future. Would there be a way to just use `nilearn.plotting.plot_glassbrain()` with `figure` argument after converting the NDVar data to a nifty image? Or do we not get access to the necessary properties to update the data?

Something minor, there should be no whitespace between function name and `(`.

proloyd:
Hi, Christian,

I just pushed a few styling fixes. 

By the way, I don't think if we need to check fsaverage at this point. What if someone morphs fsaverage to individual subjects and tries to use glassbrain?

christianbrodbeck:
> What if someone morphs fsaverage to individual subjects and tries to use glassbrain?

Since the brain is moved to the individual subject, it will not be in MNI space anymore.

proloyd:
Hi Christian, I only used _get_colorbar_and_data_ranges() private function. I think we will write our version of this function later. Directly using plotting.glassbrain() function is kinda messy and is not that flexible to update/ change. So, I did the plotting in the current way, gives you full control over all aspects of the figure. Also do you remember the unlinking issues I mentioned before? That also came because I was using the plotting.glassbrain that time.

christianbrodbeck:
There are also 
```
from nilearn._utils.numpy_conversions import as_ndarray
from nilearn._utils.niimg import _safe_get_data
from nilearn._utils.extmath import fast_abs_percentile
```
(`_utils` is private)

christianbrodbeck:
And yeah, let's not use `plot_glassbrain` then

proloyd:
I just wrote my functions (in ._nifty_utils) to reduce dependency on these private libraries. I hope you have some nice function for editing colorbar, we can use that for editing the colorbar later on.  

By the way, then shall we add fsaverage check? For that shall I just check the name in the source?

christianbrodbeck:
> I just wrote my functions (in ._nifty_utils) to reduce dependency on these private libraries. 

great :)

> I hope you have some nice function for editing colorbar, we can use that for editing the colorbar later on.

Would it be possible to add the `ColorBarMixin` to the figure? Does not have to be in this PR though, we can save it for later.   

> By the way, then shall we add fsaverage check? For that shall I just check the name in the source?

Yeah I'm not sure how best to check... for now maybe just check for the name yeah. In the future it might be possible to check whether the MRI is scaled, and if it is just load the fsaverage source space instead of the scaled one.

eelbrain/plot/_glassbrain.py
```diff
@@ -0,0 +1,430 @@
...
+
+        if ndvar:
+            if ndvar.has_case:
+                ndvar = ndvar.mean('case')
+            # instead check ndvar
+            if ndvar.has_dim('space'):
+                ndvar = ndvar.norm('space')
+
+            self._ndvar = ndvar
+            if ndvar.source.subject == 'fsaverage' and not mni_correction:
```
christianbrodbeck: What about setting `mne_correction` to `None` by default, and then here 
```
if mni_correction is None:
    mni_correction = ndvar.source.subject == 'fsaverage'

christianbrodbeck: (Then you would not have to remember to set this argument whenever plotting fsaverage)

christianbrodbeck:
Is it ready to be merged from your end?

proloyd:
yeah! I tested the final version, and it is working fine. :)

christianbrodbeck:
Thanks @proloyd !

PR #30, opened 2018-12-09
Adding draw arrow feature for glassbrain

proloyd:
Now glassbrain accepts ``draw_arrows`` argument, which if set to ``True`` arrows pointing the direction of activation is drawn over the glassbrain plots. Naturally, for this to work ``NDVar`` needs to contain space dimension (i.e 3D vector data). By default it is set to ``False``.


eelbrain/_stats/testnd.py
```diff
@@ -1866,6 +1866,28 @@ def _repr_test_args(self):
...
+        if not 1 >= p > 0:
+            raise ValueError(f"pmin={pmin}: needs to be between 1 and 0")
+        if p == 1:
+            if self._cdist.kind != 'cluster':
+                raise ValueError(f"p=1 is only a valid mask for threshold-based cluster tests")
+            mask = self._cdist.cluster_map == 0
+        else:
+            mask = self.p > p
+        mask = self._cdist.uncrop(mask, self.difference.norm('space'), True)
+        return self.difference.norm('space').mask(mask), self.difference * (self.p <= p)
```
christianbrodbeck: This function should not be changed to maintain consistency with other NDTests. If it just returns the vector difference (as the superclass method should do) you should just be able to take the norm afterwards, no?

eelbrain/plot/_base.py
```diff
@@ -666,11 +669,19 @@ def butterfly_data(
...
+        else:
+            z = y
+        bfly_data = [z.sub(source=hemi, name=hemi.capitalize()) for hemi in hemis]
+        brain_data = y
+    elif kind == 'vector':
         if resample_:
             y = resample(y, resample_, window='hamming')
-        bfly_data = [y.sub(source=hemi, name=hemi.capitalize()) for hemi in hemis]
+            z = resample(z, resample_, window='hamming')      
+        bfly_data = [z.sub(source=hemi, name=hemi.capitalize()) for hemi in hemis]
```
christianbrodbeck: I think this `butterfly_data()` is also used for surface source spaces, and there we would need norm data – you'll probably need a parameter that says whether `butterfly_data()` should return vector data if available.

eelbrain/plot/_glassbrain.py
```diff
@@ -154,7 +161,12 @@ class GlassBrain(TimeSlicerEF, ColorBarMixin, EelFigure):
     _make_axes = False
     _display_time_in_frame_title = True
 
-    def __init__(self, ndvar, cmap=None, vmin=None, vmax=None, dest='mri', mri_resolution=False, mni305=None, black_bg=False, display_mode=None, threshold=None, colorbar=False, draw_cross=True, annotate=True, alpha=0.7, plot_abs=False, symmetric_cbar="auto", interpolation='nearest', **kwargs):
+    def __init__(self, ndvar, cmap=None, vmin=None, vmax=None, dest='mri',
+                 draw_arrows=False, mri_resolution=False, mni305=None,
+                 black_bg=False, display_mode=None, threshold='auto',
+                 colorbar=False, draw_cross=True, annotate=True,
+                 alpha=0.7, plot_abs=False, symmetric_cbar="auto",
+                 interpolation='nearest', **kwargs):
```
christianbrodbeck, line 167: I would say `draw_arrows` should be true by default, and it should be disabled when `ndvar` is scalar. Or do you think there will be more cases where we want to plot vector data without arrows?

eelbrain/plot/_glassbrain.py
```diff
@@ -291,11 +340,94 @@ def __init__(self, ndvar, cmap=None, vmin=None, vmax=None, dest='mri', mri_resol
...
+                self.display.axes[axis].ax.images.clear()
+
+    # used by _update_time
+    def _remove_arrows(self):
+        for q in self._quivers:
+            q.remove()
+
+    def _remove_arrows(self):
+        for q in self._quivers:
+            q.remove()
```
christianbrodbeck: Duplicate method?

eelbrain/plot/_glassbrain.py
```diff
@@ -327,13 +462,12 @@ def _colorbar_params(self):
...
 
     @classmethod
-    def butterfly(
-            cls, y, cmap=None, vmin=None, vmax=None,
-            dest='mri', mri_resolution=False, mni305=None,
-            black_bg=False, display_mode=None, threshold=None,
-            colorbar=False, alpha=0.7, plot_abs=False,
-            symmetric_cbar="auto", interpolation='nearest',
-            w=5, h=2.5, xlim=None, name=None, **kwargs):
+    def butterfly(cls, y, cmap=None, vmin=None, vmax=None, draw_arrows=False,
```
christianbrodbeck: The order of arguments should be consistent with `__init__`

eelbrain/plot/_glassbrain.py
```diff
@@ -388,6 +522,11 @@ def butterfly(
             different colors. See `examples <http://nilearn.github.io/auto_examples/
             01_plotting/plot_demo_glass_brain_extensive.html>`_. Only affects
             GlassBrain plot.
+        draw_arrows: boolean
```
christianbrodbeck, line 521: The order in the documentation should be same as in the function signature

eelbrain/plot/_glassbrain.py
```diff
@@ -597,3 +740,46 @@ def _fast_abs_percentile(ndvar, percentile=80):
...
+    img: Nifti image/object
+        Image to get data.
+    Returns
+    -------
+    data: numpy array
+        get_data() return from Nifti image.
+        # inspired from nilearn._utils.niimg._safe_get_data
+    """
+    data = img.get_data()
+    non_finite_mask = np.logical_not(np.isfinite(data))
```
christianbrodbeck: Do you really need this check? Since this happens in the animation, it might be more efficient to guarantee that the data is ok when generating it?

christianbrodbeck (changes requested):
Look good already! Some potential changes to consider, let me know if something is not clear. Thanks!

eelbrain/plot/_glassbrain.py
proloyd: Done!

eelbrain/plot/_glassbrain.py
proloyd: Thanks for pointing this out!

eelbrain/plot/_glassbrain.py
proloyd: Good point, I bypassed this check by setting a global check while converting ndvar to nifti/ volume format.

eelbrain/plot/_glassbrain.py
proloyd: I did not get this comment. I just made a separate function so that readability improves. 
Also, if we decide to add this feature on GUI in future, then it will be helpful in key-binding.

eelbrain/plot/_glassbrain.py
proloyd: I made the changes.

eelbrain/plot/_base.py
proloyd: Okay! Made the changes. https://github.com/proloyd/Eelbrain/blob/3d538b0c29bbac23f71749a99b23e076cc60184f/eelbrain/plot/_base.py#L667

eelbrain/_stats/testnd.py
proloyd: Now rolled back to previous version of the function. Actually the np.ma.masked_array function does not do broadcasting, so did that. https://github.com/proloyd/Eelbrain/blob/e08725268dc4327454f5e0994d175c98a4da90f5/eelbrain/_data_obj.py#L3993

eelbrain/plot/_glassbrain.py
proloyd: Oh..I see now. Cool. I dont know how this got duplicated.

proloyd (commented):
np.ma.maskedarray does not support broadcasting, so had to manually do it.

eelbrain/plot/_base.py
```diff
@@ -666,12 +667,26 @@ def butterfly_data(
     if kind == 'ndvar':
         if y.has_case:
             y = y.mean('case')
-        if y.has_dim('space'):
-            y = y.norm('space')
         if resample_:
             y = resample(y, resample_, window='hamming')
+        if y.has_dim('space'):
```
proloyd, line 672: taking norm does not preserve the mask, so added this few lines to take care of that.

eelbrain/_data_obj.py
```diff
@@ -3990,6 +3990,9 @@ def mask(self, mask, name=None):
         x_mask = self._ialign(mask)
         if x_mask.dtype.kind != 'b':
             x_mask = x_mask.astype(bool)
+        # Masked array creation does not support broadcasting
```
proloyd: Apparently , masked array creation does not support broadcasting, so needed to mask three directions separately.

eelbrain/plot/_glassbrain.py
```diff
@@ -199,20 +210,55 @@ def __init__(self, ndvar, cmap=None, vmin=None, vmax=None, dest='mri', mri_resol
             else:
                 img0 = img
                 imgs = time = t0 = None
+            if draw_arrows:
+                if not ndvar.has_dim('space'):
+                    warnings.warn('Cannot draw arrows:'
+                                  'ndvar does not have space dimension.'
+                                  'Continuing without arrows...')
```
christianbrodbeck: I would skip the warning (especially now that `draw_arrows` is on by default)

eelbrain/plot/_base.py
```diff
@@ -624,6 +624,7 @@ def butterfly_data(
         hemi: str,
         resample_: int = None,
         colors: bool = False,
+        return_vector_data = False,
```
christianbrodbeck: To be consistent with the rest of the function signature you could add the type specification (`: bool`)

eelbrain/plot/_base.py
christianbrodbeck: Oh but it should :) do you want to look into fixing it? I can also look into it if it looks to complex.

eelbrain/_data_obj.py
christianbrodbeck: Hmm, according to the documentation it does not handle broadcasting. If it were to it should handle it in a general way, but that might be tricky. Say, the NDVar has shape `(3, 3)`, and `.mask()` is called with a shape `(3,)` array, which dimension do you broadcast? It could assume always the first dimension? But probably easier would be to broadcast the mask in `NDDifferenceTest.masked_difference` with a check like `if isinstance(self, Vector)` (that will also be True for subclasses of `Vector`)

eelbrain/plot/_glassbrain.py
```diff
@@ -432,8 +571,9 @@ def butterfly(
             get_app().jumpstart()
 
         # GlassBrain plot
-        p_glassbrain = GlassBrain(brain_data, cmap, vmin, vmax, dest, mri_resolution, mni305, black_bg, display_mode, threshold, colorbar, True, True, alpha, plot_abs, symmetric_cbar, interpolation, h=h, name=name, **kwargs)
-
+        p_glassbrain = GlassBrain(brain_data, cmap, vmin, vmax, dest, draw_arrows, mri_resolution, mni305,
```
christianbrodbeck: Looks like `draw_arrows` is now in the wrong place here

eelbrain/plot/_glassbrain.py
christianbrodbeck: But looks like it changed for `__init__` since then :)

eelbrain/plot/_base.py
proloyd: Cool, Doing that. By the way, I was not familiar with this type of function signature. Is it some special function signature type?

eelbrain/plot/_base.py
proloyd: No, it should not. Like you said previously, suppose your original array a was (3, 3), and you took norm(a), then how will it decide the mask for that scalar? Same applies when it's a 1D (3, ) array. Right? I though of adding this `mask`ed `norm` in the data_obj, but it does not make any sense.

eelbrain/plot/_base.py
christianbrodbeck: https://docs.python.org/3.6/library/typing.html

eelbrain/plot/_base.py
proloyd: I guess this only the place, where we will use masking the norm from the vectors, so it makes sense to put it over here.

christianbrodbeck: It should behave like `numpy` does. If you're worried about the norm of a vector for which only some elements are masked, we could raise an error for that for now. For the significance case, it should always be either all or none of the elements.

eelbrain/plot/_glassbrain.py
proloyd: Okay. Doing that.

eelbrain/_data_obj.py
proloyd: Implemented that.

eelbrain/plot/_glassbrain.py
```diff
@@ -199,20 +210,52 @@ def __init__(self, ndvar, cmap=None, vmin=None, vmax=None, dest='mri', mri_resol
...
+                    for direction in ndvar.space._directions:
+                        dir_img = _stc_to_volume(ndvar.sub(space=direction), src,
+                                                 dest, mri_resolution, mni305)
+                        if ndvar.has_dim('time'):
+                            dir_imgs.append([index_img(dir_img, i)
+                                             for i in range(len(ndvar.time))])
+                        else:
+                            dir_imgs.append([dir_img])
+                    if plot_abs:
+                        warnings.warn('Cannot use maximum intensity projection'
```
christianbrodbeck: I would just raise a `ValueError` if parameters are inconsistent/impossible.

christianbrodbeck: Does not seem to handle multiline deletion
```suggestion
                        raise ValueError(f"Cannot use plot_abs={plot_abs} with draw_arrows={draw_arrows}")
```

eelbrain/plot/_glassbrain.py
```diff
@@ -493,11 +622,18 @@ def _stc_to_volume(ndvar, src, dest='mri', mri_resolution=False, mni305=False):
...
     src_type = src[0]['type']
     if src_type != 'vol':
         raise ValueError(f"You need a volume source space. Got type: {src_type}")
 
+    # take care of masked array.
+    if isinstance(ndvar.x, np.ma.MaskedArray):
+        ndvar = ndvar.copy('copy')
+        ndvar.x[ndvar.x.mask] = 0
+        ndvar.x = ndvar.x.data
+
```
christianbrodbeck: No need for this, it's handled in get_data (https://eelbrain.readthedocs.io/en/latest/generated/eelbrain.NDVar.get_data.html#eelbrain.NDVar.get_data)

proloyd: Which one did you pointed here? The ndvar.copy('copy') or the Note? 
ndvar.copy('copy') was needed here, otherwise `ndvar.x[ndvar.x.mask] = 0` was changing the original array.
I should have removed the Note, btw.

eelbrain/plot/_base.py
proloyd: Actually, numpy norm does not acre about the mask. It just returns a vanilla ndarray, removing all masking info. So, I tried to add this masking thing in  ndvar.norm but it does not make sense in the case  when some elements along the axis over which norm is taken are masked, not all. So, I just moved this code in this position.

eelbrain/plot/_glassbrain.py
proloyd: Done.

eelbrain/plot/_glassbrain.py
proloyd: Oh, I now got it. You meant the masking. Taking care of it.

eelbrain/_data_obj.py
```diff
@@ -4107,6 +4107,12 @@ def norm(self, dim, ord=2, name=None):
         x = norm(self.x, ord, axis)
         if self.ndim == 1:
             return x
+        if isinstance(self.x, np.ma.masked_array):
+            if dim == 'space':
+                mask = np.rollaxis(self.x.mask, self.get_axis('space'))[0]
+                x = np.ma.masked_array(x, mask)
+            else:
+                raise NotImplementedError(f"masked data norm along dim={dim}")
```
proloyd: Until we figure out a sensible way of returning norm of masked data, attempting 'norm' for masked data along any dim except space will raise error.

eelbrain/plot/_base.py
proloyd: I moved this to ndvar.norm. Let me know if that makes sense.

eelbrain/_data_obj.py
christianbrodbeck: What about 

```
all_masked = np.all(self.x.mask, axis)
any_masked = np.any(self.x.mask, axis)
if np.any(all_masked != any_masked):
    raise ValueError(f"Norm along {dim!r} with inconsistent mask")
mask = all_masked
```

christianbrodbeck: What about 
```
all_masked = np.all(self.x.mask, axis)
any_masked = np.any(self.x.mask, axis)
if np.any(all_masked != any_masked):
    raise ValueError(f"Norm along {dim!r} with inconsistent mask")
mask = all_masked
```

eelbrain/_data_obj.py
```diff
@@ -4107,6 +4107,13 @@ def norm(self, dim, ord=2, name=None):
         x = norm(self.x, ord, axis)
         if self.ndim == 1:
             return x
+        if isinstance(self.x, np.ma.masked_array):
```
proloyd, line 4110: great idea! But I guess this will almost always raise error if the dim is not `space`.

christianbrodbeck: That's fine, if someone has a use case that makes sense it can be modified. There might also be other cases where this would work, for example, in sensor space data you could mask a specific sensor.

proloyd: Yeah! Sure. Haven't thought of that.

eelbrain/plot/_base.py
```diff
@@ -666,12 +667,18 @@ def butterfly_data(
     if kind == 'ndvar':
         if y.has_case:
             y = y.mean('case')
-        if y.has_dim('space'):
```
christianbrodbeck, line 669: So currently I think this function won't work as expected for `Vector` tests, right? There, `y = data.masked_difference()` will be vector data, whereas significance will not. But I can look into that if you'd rather not :) If you do, you could also add some tests to test the new features in `test_glassbrain.py`.

eelbrain/plot/_glassbrain.py
```diff
@@ -37,6 +37,8 @@
 """
 import warnings
 
+from nilearn.image.resampling import get_bounds
```
christianbrodbeck, line 40: I think there was a problem with `nilearn` imports at module level because it imports `pyplot` immediately, messing up the matplotlib backend situation

eelbrain/plot/_base.py
christianbrodbeck: Ah I realize it's tested already because `draw_arrows` is on by default:)

eelbrain/plot/_glassbrain.py
```diff
@@ -154,7 +160,11 @@ class GlassBrain(TimeSlicerEF, ColorBarMixin, EelFigure):
     _make_axes = False
     _display_time_in_frame_title = True
 
-    def __init__(self, ndvar, cmap=None, vmin=None, vmax=None, dest='mri', mri_resolution=False, mni305=None, black_bg=False, display_mode=None, threshold=None, colorbar=False, draw_cross=True, annotate=True, alpha=0.7, plot_abs=False, symmetric_cbar="auto", interpolation='nearest', **kwargs):
+    def __init__(self, ndvar, cmap=None, vmin=None, vmax=None, dest='mri',
```
christianbrodbeck, line 163: Can you keep the previous default arguments?

eelbrain/plot/_base.py
proloyd: Yeah, I took care of that already in vector test. :)

eelbrain/plot/_glassbrain.py
proloyd: yeah I remeber, but this time nothing happened, it passed the tests without any hiccups. If you want I can move it inside glassbrain.

proloyd: Actually, nilearn.image does not import pyplot, so there is no chance of messing up backend. :)

eelbrain/plot/_base.py
proloyd: By the way, I am not sure how to add tests for the new features. Any hint on that?

eelbrain/plot/_glassbrain.py
proloyd: Sure, I changed it here, and also for the `butterfly` method.

eelbrain/plot/_glassbrain.py
```diff
@@ -313,10 +441,13 @@ def _update_time(self, t, fixate):
         # A little hack to make sure that the display
         # still has correct colorbar flag.
         self.display._colorbar = self.colorbar
+
+        # take care of arrows
+        if self._arrows: self._remove_arrows()
+        if self._arrows: self._add_arrows(index)
```
christianbrodbeck: This style is discouraged, it's clearer to just write 

```
if x:
    do_a()
    do_b()
```

eelbrain/plot/_glassbrain.py
```diff
@@ -388,6 +517,10 @@ def butterfly(
             different colors. See `examples <http://nilearn.github.io/auto_examples/
             01_plotting/plot_demo_glass_brain_extensive.html>`_. Only affects
             GlassBrain plot.
+        draw_arrows: boolean
+            Draw arrows in the direction of activation over the glassbrain plots. Naturally, for this to work
+            ``ndvar`` needs to contain space dimension (i.e 3D vectors). By default it is set to
+            (``True``).
```
christianbrodbeck: The parentheses around True seem off.

christianbrodbeck (commented):
Ok, just some stylistic nitpicking, I'll make a PR to show you how to go about testing

eelbrain/plot/_glassbrain.py
christianbrodbeck: Yeah I think you're right, fine to leave it then

eelbrain/plot/_glassbrain.py
proloyd: I see. Will take a note of that.

eelbrain/plot/_glassbrain.py
proloyd: Okay. I guess I fixed it

eelbrain/plot/_glassbrain.py
```diff
@@ -388,6 +518,10 @@ def butterfly(
             different colors. See `examples <http://nilearn.github.io/auto_examples/
             01_plotting/plot_demo_glass_brain_extensive.html>`_. Only affects
             GlassBrain plot.
+        draw_arrows: boolean
+            Draw arrows in the direction of activation over the glassbrain plots.
+            Naturally, for this to work ``ndvar`` needs to contain space dimension
+            (i.e 3D vectors). By default it is set to (``True``).
```
christianbrodbeck: Shouldn't it be this?
```suggestion
            (i.e 3D vectors). By default it is set to ``True``.
```

proloyd: Actually in plot_abs it was like (``False``), so I used the same. Changing it then. :)

eelbrain/plot/tests/test_glassbrain.py
```diff
@@ -1,20 +1,36 @@
...
+    p = plot.GlassBrain(ndvar_30, draw_arrows=False)
     p.close()
 
     # time series
-    p = plot.GlassBrain(ndvar.sub(time=0.030), show=False)
+    p = plot.GlassBrain(ndvar)
+    p.set_time(.03)
+    p.close()
+
+    # masked data
```
proloyd, line 27: I added this test for masked data. Is it cumbersome to create the masked array like this? In case you find this unfit we can remove this part.  

Btw thanks for letting me know how to implement this tests. So far I was testing this manually and that was a pain. :( Thank you very much for the guide. :)

eelbrain/plot/tests/test_glassbrain.py
```diff
@@ -1,20 +1,36 @@
...
+    import numpy as np
+    h = ndvar.sub(time=0.030)
+    c = _fast_abs_percentile(h)
+    mask = h.norm('space') < c
+    mask_x = np.repeat(h._ialign(mask), 3, h.get_axis('space'))
+    mask = NDVar(mask_x, h.dims)
+    y = h.mask(mask)
+    p = plot.GlassBrain(y, show=False)
```
christianbrodbeck: You should not need `show=False`, that should be implied by the `@hide_plots` decorator
```suggestion
    p = plot.GlassBrain(y)
```

eelbrain/plot/tests/test_glassbrain.py
christianbrodbeck: Yes great. You could in principle hard code parameters like `c` because the data will always be the same for the test, but it's a good test to add.

christianbrodbeck (approved):
Let me know when it's ready to be merged from your end!

eelbrain/plot/tests/test_glassbrain.py
proloyd: Cool!

eelbrain/plot/tests/test_glassbrain.py
proloyd: Cool. Now incorporated that. :)

proloyd:
Once the appveyor build is passed, the branch will be ready to be merged. It is taking so long. I dunno why?

christianbrodbeck:
Great, Thanks!

proloyd:
My pleasure!

PR #31, opened 2019-01-09
Adding Hotelling-T ** 2 statistic in testnd.Vector.

proloyd:
Info: The T ** 2 statistic computation happens fully in Cython (vector.pyx).


christianbrodbeck:
Is your impression that the *t*-version of the test should be the default?

proloyd:
Yes, It should be default. But since there are other NDDifferenceTest (which I did not get to touch yet) still using vector-norm. So, to be consistent I kept it as an option. When I get to use those tests, hopefully I will also change them, then we can make T**2 version of Vector test the default!

eelbrain/_stats/testnd.py
```diff
@@ -1853,12 +1855,18 @@ def __init__(self, y, match=None, sub=None, ds=None,
         v_dim = ct.y.dimnames[cdist._vector_ax + 1]
         v_mean = ct.y.mean('case')
         v_mean_norm = v_mean.norm(v_dim)
-        cdist.add_original(v_mean_norm.x if v_mean.ndim > 1 else v_mean_norm)
+        if use_t2_stat:
+            t2_map = self._vector_t2_map(ct.y)
+            cdist.add_original(t2_map.x if v_mean.ndim > 1 else v_mean_norm)
```
christianbrodbeck:
```suggestion
            cdist.add_original(t2_map.x if v_mean.ndim > 1 else t2_map)
```

This is because if you run the test with a variable where the only dimension other than case is space, then the output of the test function should be a scalar

eelbrain/_stats/tests/test_testnd.py
```diff
@@ -638,7 +638,7 @@ def test_vector():
     # single vector
     ds = datasets.get_uv(vector=True)
     res = testnd.Vector('v[:40]', ds=ds, samples=10)
-    assert res.p == 0.1
+    assert res.p == 0.0
```
christianbrodbeck, line 641: Hmmm do you know why you had to change this? I.e. if 0.1 failed for you, why did it pass on `master`?

christianbrodbeck (commented):
Looks great, just a few questions. Maybe you could also add a single vector test? (Just duplicate one of the tests at the top of `test_vector` for `use_t2_stat`?)

eelbrain/_stats/tests/test_testnd.py
proloyd: I don't know exactly, But I repeated that test several times (100, may be more) it gave me 0.0 always.

eelbrain/_stats/testnd.py
proloyd: By the way, it took me a while to figure out there exists np scalar arrays, and how to declare one. :(

christianbrodbeck: Yeah they caused me some headaches too when I first encountered them :)

eelbrain/_stats/tests/test_testnd.py
christianbrodbeck: Ooh I see, it must be because of your redefinition of `rotation_matrices()`.

christianbrodbeck: I think there might be some problem with this. Running the following code, it seems to me that the sampling is biased towards two poles (it's not in `master`):

```
from mpl_toolkits.mplot3d import Axes3D  # noqa: F401 unused import
import matplotlib.pyplot as plt
from eelbrain._stats.testnd import *
n_cases = 2000
phi = np.random.uniform(0, 2 * pi, n_cases)
theta = np.arccos(np.random.uniform(-1, 1, n_cases))
rotation = vector.rotation_matrices(phi, theta, np.empty((n_cases, 3, 3)))
origin = np.array([1, 0, 0]).reshape((1, 3))
vecs = [origin.dot(r) for r in rotation]
vecs = np.vstack(vecs)
fig = plt.figure()
ax = fig.add_subplot(111, projection='3d')
ax.scatter(*vecs.T)
ax.set_xlabel('X Label')
ax.set_ylabel('Y Label')
ax.set_zlabel('Z Label')
plt.show()
```

Can you reproduce this? Trying to figure out why I realize it would be good to document the procedure better :) I noticed that compared to http://mathworld.wolfram.com/SpherePointPicking.html I switched theta and phi (in the Python code at `_vector_mean_norm_perm()`), not sureif that has something to do with it?

christianbrodbeck:
> Yes, It should be default. But since there are other NDDifferenceTest (which I did not get to touch yet) still using vector-norm.

You could try to add it to `VectorDifferenceRelated`, because it's based on a test of the difference that is equivalent to `Vector`. I think it would be fine to make it default, could even add the argument to the independent test and raise a `NotImplementedError`. I will definitely add a note that the tests are experimental and might change in the future.

eelbrain/_stats/tests/test_testnd.py
proloyd: The sampled points were indeed random (uniformly distributed over the unit sphere), but the rotation matrix we created (both of us) was not doing random rotation. :(. To cover the whole space in `3D` , you have to have three parameters (`3` angles for example), but in the rotation matix we had only `2` (`theta` and `phi`). The clear picture of rotation in `3D` can be found: [here](https://en.wikipedia.org/wiki/Rotation_matrix)
and [here](http://www.euclideanspace.com/maths/geometry/rotations/conversions/angleToMatrix/).
Right now, I am testing with them. Once I am confident, I will push the changes.
Thanks for the piece of the code though, otherwise I could not have caught this.

christianbrodbeck: Great, thanks!

proloyd: Hi, 
I just pushed the commit. Here is the piece of code that will help you visualization. you can try changing the ``origin`` point:
```
from mpl_toolkits.mplot3d import Axes3D  # noqa: F401 unused import
import matplotlib.pyplot as plt
from eelbrain._stats.testnd import *
n_cases = 2000
angles = np.random.uniform(0, 2 * pi, (2, n_cases))
theta = angles[0]
phi = np.arccos(np.random.uniform(-1, 1, n_cases))
xi = angles[1]
rotation_points = vector.rotation_points(phi, theta, np.empty((n_cases, 3)))
rotation = vector.rotation_matrices(phi, theta, xi, np.empty((n_cases, 3, 3)))
origin = np.array([0.3, 0.3, 1]).reshape((1, 3))
vecs = [origin.dot(r.T) for r in rotation]
vecs = np.vstack(vecs)
fig = plt.figure()
ax = fig.add_subplot(121, projection='3d')
ax.scatter(*vecs.T)
ax.set_xlabel('X Label')
ax.set_ylabel('Y Label')
ax.set_zlabel('Z Label')
ax = fig.add_subplot(122, projection='3d')
ax.scatter(*rotation_points.T)
ax.set_xlabel('X Label')
ax.set_ylabel('Y Label')
ax.set_zlabel('Z Label')
plt.show() 
```
With the new rotation matrix, you can now cover the whole sphere!

christianbrodbeck: I think there'e still a problem 😔 Try the script below, it suggests the density is much bigger on the `x=1` end than on the `x=-1` end ...

```python
import numpy as np
from mpl_toolkits.mplot3d import Axes3D
import matplotlib.pyplot as plt
from eelbrain._stats.testnd import *

n_cases = 2000
phi = np.random.uniform(0, 2 * pi, n_cases)
theta = np.arccos(np.random.uniform(-1, 1, n_cases))
xi = np.random.uniform(0, 2 * pi, n_cases)
rotation = vector.rotation_matrices(phi, theta, xi, np.empty((n_cases, 3, 3)))
origin = np.array([1, 0, 0]).reshape((3, 1))
vecs = [r.dot(origin).T for r in rotation]
vecs = np.vstack(vecs)
x, y, z = vecs.T

# plot
# fig = plt.figure(figsize=(5, 5))
# ax = fig.add_subplot(111, projection='3d')
# ax.scatter(x, y, z)
# ax.set_xlabel('X Label')
# ax.set_ylabel('Y Label')
# ax.set_zlabel('Z Label')
# plt.show()

print(np.histogram(x, range=(-1, 1))[0])
```

christianbrodbeck: Hey, it seems a little silly to only think of that now, so maybe I'm missing something, but: if we want to rotate randomly with uniform distribution on the sphere, then the original direction of the vector doesn't actually matter! So there's no need for rotation... We can just generate random directions on the unit sphere (using phi and theta) and scale them with the length of the original vector, right?

proloyd: Ha ha, I thought of this, but it sounds silly. After generating the directions, just multiplying it with the norm of the vector should do the trick. 😁 But it seems like a little hack. 

Fortunately, I figured out how to generate random rotation matrices and will pull these changes soon, once I am done with testing. 

basically instead of randomly sampling xi, we have to use another distribution. You can check via the following code:
``` from mpl_toolkits.mplot3d import Axes3D  # noqa: F401 unused import
import matplotlib.pyplot as plt
from eelbrain._stats.testnd import *

import numpy as np

def rejection_sampling_xi(iter=1000):
    """Sample from p(x) = 2 * np.sin(x/2) ** 2 / pi"""
    from math import pi
    samples = []

    while True:
        z = np.random.uniform(0, pi)
        u = np.random.uniform(0, 2/pi)

        if u <= 2 * np.sin(z/2) ** 2 / pi:
            samples.append(z)
            if len(samples) == iter:
                break

    return np.array(samples)

def rotation_points(phi, theta):
    raxis_x = np.sin(phi) * np.cos(theta)
    raxis_y = np.sin(phi) * np.sin(theta)
    raxis_z = np.cos(phi)
    return np.array((raxis_x, raxis_y, raxis_z))


n_cases = 5000
angles = np.random.uniform(0, 1, (3, n_cases))
theta = 2 * pi * angles[0]
phi = np.arccos(2 * angles[1] - 1)
xi = rejection_sampling_xi(n_cases)
rotation_points = rotation_points(phi, theta)
rotation = vector.rotation_matrices(phi, theta, xi, np.empty((n_cases, 3, 3)))
origin = np.array([0, 1, 0]).reshape((1, 3))
vecs = [origin.dot(r.T) for r in rotation]
vecs = np.vstack(vecs)
x, y, z = vecs.T


# # plot
# fig = plt.figure()
# ax = fig.add_subplot(121, projection='3d')
# ax.scatter(*vecs.T)
# ax.set_xlabel('X Label')
# ax.set_ylabel('Y Label')
# ax.set_zlabel('Z Label')
# # plt.show()
# ax = fig.add_subplot(122, projection='3d')
# ax.scatter(*rotation_points)
# ax.set_xlabel('X Label')
# ax.set_ylabel('Y Label')
# ax.set_zlabel('Z Label')
# plt.show()
print(np.histogram(x, range=(-1, 1))[0])
```

christianbrodbeck: Ah great! I realized using only random directions would not work because we want to keep relative directions within each case. Otherwise we would destroy the spatio-temporal covariance pattern...

eelbrain/_stats/testnd.py
```diff
@@ -1890,16 +1900,55 @@ def _repr_test_args(self):
...
+        return vector.t2_stat_rotated(y, rotation, out)
+
+    @staticmethod
+    def _vector_t2_map(y):
+        ndim = len(y.dimnames)
+        dimnames = ('case', 'space',) + (None,) * (ndim - 2)
+        dimnames = y.get_dimnames(dimnames)
+        x = y.get_data(dimnames)
+        if len(dimnames) == 2:
+            t2_map = stats.t2_1samp(x,)
```
christianbrodbeck: The comma does not do anything, so this line could be moved out of the conditional

eelbrain/_stats/tests/test_testnd.py
christianbrodbeck: Still uneven 😓

```
from math import pi
import numpy as np
from eelbrain._stats.testnd import *


def _vector_mean_norm_perm(n_cases=2000):
    phi = np.random.uniform(0, 2 * pi, n_cases)
    theta = np.arccos(np.random.uniform(-1, 1, n_cases))
    xi = Vector._rejection_sampling_xi(n_cases)
    return vector.rotation_matrices(phi, theta, xi, np.empty((n_cases, 3, 3)))

# source vector
rotation = _vector_mean_norm_perm()
origin = np.array([1, 0, 0]).reshape((3, 1))
vecs = [r.dot(origin).T for r in rotation]
vecs = np.vstack(vecs)
x, y, z = vecs.T

print(np.histogram(x, range=(-1, 1))[0])
```

gives me arrays like
```
[386 293 262 217 198 174 156 146 119  49]
```

proloyd: Actually, there is a problem with the peace of code in above. You exchanged the role of phi and theta. I also saw that I did the same mistake in the `Vector` methods. :( I just made them right. I will push the changes after some testing. In the mean time, try running this code.
```
from math import pi
import numpy as np
from eelbrain._stats.testnd import *


def _vector_mean_norm_perm(n_cases=2000):
    phi = np.arccos(np.random.uniform(-1, 1, n_cases))
    theta = np.random.uniform(0, 2 * pi, n_cases)
    xi = Vector._rejection_sampling_xi(n_cases)
    return vector.rotation_matrices(phi, theta, xi, np.empty((n_cases, 3, 3)))

# source vector
rotation = _vector_mean_norm_perm()
origin = np.array([1, 0, 0]).reshape((3, 1))
vecs = [r.dot(origin).T for r in rotation]
vecs = np.vstack(vecs)
x, y, z = vecs.T

print(np.histogram(x, range=(-1, 1))[0])
```

eelbrain/_stats/testnd.py
```diff
@@ -1890,16 +1900,55 @@ def _repr_test_args(self):
...
-        phi = np.random.uniform(0, 2 * pi, n_cases)
-        theta = np.arccos(np.random.uniform(-1, 1, n_cases))
-        rotation = vector.rotation_matrices(phi, theta, np.empty((n_cases, 3, 3)))
+        phi = np.arccos(np.random.uniform(-1, 1, n_cases))
+        theta = np.random.uniform(0, 2 * pi, n_cases)
+        xi = self._rejection_sampling_xi(n_cases)
+        rotation = vector.rotation_matrices(phi, theta, xi, np.empty((n_cases, 3, 3)))
         return vector.mean_norm_rotated(y, rotation, out)
 
+    def _vector_t2_map_perm(self, y, out, seed):
```
christianbrodbeck: I would refactor this so that there is less code duplication between this and the previous method. Since the methods are getting more complex, calling each other, it might also make more sense to refactor them into module level functions...

eelbrain/_stats/testnd.py
```diff
@@ -1890,16 +1900,55 @@ def _repr_test_args(self):
...
+            return np.float64(t2_map)
+        else:
+            t2_map = stats.t2_1samp(x)
+            dims = (y.get_dim(dimnames[i]) for i in range(2, ndim))
+            return NDVar(t2_map, dims)
+
+    @staticmethod
+    def _rejection_sampling_xi(iter=1000):
+        """Sample from p(x) = 2 * np.sin(x/2) ** 2 / pi"""
+        from math import pi
```
christianbrodbeck: `pi` is already imported at module level

eelbrain/_stats/testnd.py
```diff
@@ -1980,11 +2032,19 @@ def __init__(self, y, x, c1=None, c0=None, match=None, sub=None, ds=None,
...
+        if use_t2_stat:
+            t2_map = self._vector_t2_map(ct.y)
+            cdist.add_original(t2_map.x if v_mean.ndim > 1 else t2_map)
+        else:
+            cdist.add_original(v_mean_norm.x if v_mean.ndim > 1 else v_mean_norm)
 
         if cdist.do_permutation:
             iterator = random_seeds(samples)
-            run_permutation(self._vector_mean_norm_perm, cdist, iterator)
+            iterator = random_seeds(samples)
```
christianbrodbeck: duplicate line

christianbrodbeck (commented):
Yay!

I just added some stylistic comments, let me know if you prefer me to make the suggested changes myself and make a PR into your PR :)

One issue now is that Appveyor gets stuck. Do you have an idea why this could be? It looks like it's only Appveyor and not Travis, but it's consistent over the past couple of builds...

proloyd:
This might be the issue with travis.

image

christianbrodbeck:
NewIdRef is new for wx 4, you might just have wx 3 installed. My Appveyor builds don’t fail, so I suspect it’s something more specific to the new tests… you could try commenting out the whole vector test to see if that makes appveyor pass? And then re-include specific tests to see where it hangs?

proloyd:
Finally it passed appveyor. I noticed that for last few times the check was basically stalled due to  something. Now it seems it's fixed.

eelbrain/_stats/permutation.py
```diff
@@ -235,3 +236,57 @@ def random_seeds(samples, seed=0):
...
+    samples = []
+    while True:
+        z = np.random.uniform(0, pi)
+        u = np.random.uniform(0, 2 / pi)
+
+        if u <= 2 * np.sin(z / 2) ** 2 / pi:
+            samples.append(z)
+            if len(samples) == n:
+                break
+    return np.array(samples)
```
christianbrodbeck: Since this function is used during permutation it might be somewhat performance sensitive, so here are some potential optimizations:

## 1
Since `numpy` always uses arrays, that adds a lot of overhead when working with scalars. You could use the builtin `random` module instead:
```
In [13]: timeit random.random() * pi                                                                                                                                         
219 ns ± 5.64 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
In [8]: timeit np.random.uniform(0, pi)                                                                                                                                      
2.1 µs ± 76.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Same for the `np.sin`, `math.sin` should be much faster.

## 2
Since we already know the number of elements, you could initialize an array at the beginning and then replace values in that array directly instead of constructing a list and then converting it to an array.

christianbrodbeck (commented):
Just one suggestion for optimizing the permutations, otherwise looks great!

eelbrain/plot/_glassbrain.py
```diff
@@ -396,7 +396,8 @@ def _add_arrows(self, t):
 
                     im = display_ax.ax.quiver(x, y, dir_data[0], dir_data[1],
                                               color=color,
-                                              scale=self.arrow_scale)
+                                              scale=self.arrow_scale,
+                                              )
```
christianbrodbeck: And if you could just undo this change to keep the history cleaner

proloyd: I see, I have no idea how this came. :(

christianbrodbeck: Probably from one of the times merging the `master` into the PR branch. What you could do instead is rebasing, if you're not familiar I can show you some time

proloyd: Yes, please. Some help in this frontier will be welcome. I understand a bit of rebasing now, but tracking history is still not my cup of tea, and that makes all the trouble.

proloyd: Added to that `git log` or `git diff` stopped working on my machine, just now I figured what was missing.

eelbrain/_stats/permutation.py
```diff
@@ -235,3 +236,59 @@ def random_seeds(samples, seed=0):
...
+    Returns
+    -------
+    ndarray
+        samples drawn from the distribution
+    """
+    samples = np.empty(n)
+    i = 0
+    while True:
+        z = np.random.random() * pi
+        u = np.random.random() * 2 / pi
```
christianbrodbeck: I meant to use `random.random()` from the [Python random module](https://docs.python.org/3/library/random.html#random.random) :) `np.random.random()` is still slower. That would also entail seeding it with `random.seed()` in addition to `numpy.random.seed`...

eelbrain/_stats/tests/test_testnd.py
```diff
@@ -658,7 +672,7 @@ def test_vector():
     configure(n_workers=True)
```
christianbrodbeck, line 672: You said this block caused the Appveyor problem? Did you have to change anything about it or did it resolve itself?

proloyd: It did resolve later. I tried once with it commented. From then it started to pass the test. Then eventually, I uncommented it, and it is passing the test now.

christianbrodbeck: Hm, maybe it was a temporary issue on the Appveyor platform then...

eelbrain/_stats/permutation.py
proloyd: Oh okay! I missed that.

proloyd:
Maybe!

christianbrodbeck:
Great, thanks!!!

PR #32, opened 2019-01-29
Vectorstats

proloyd:
'True' ``t**2`` test. 
(But please don't merge it now, intended for testing only.)


proloyd:
Christian, can you look at the error status in windows build once?

christianbrodbeck:
Looks like there are errors compiling `dsyevh3C`, is it supposed to compile under Windows? Have you tried compiling in Windows on your computer?

proloyd:
No. I haven't tried on my windows. Since, it's a C program, it needs to be
compiled. Doesn't it?
What I came to understand is that the default windows C compiler `cl.exe`
does not support C-standard 99, in which the C codes were written. So, it's
throwing error for the syntax used in the C program.
What I can do is:
1. Change the C codes so that they conform to C-11 standard, which `cl.exe`
supports. or,can use
2. Use gcc for windows to compile them.
Do you see any other way, like distributing the .o files etc, so that we
just have to link them, instead of compiling during installation?

By the way, the test should be working under Linux/ MacOS, so you can use
them by yourself now.

christianbrodbeck:
I don’t know enough about compiling on Windows to prefer either of these methods. I think distributing compiled (.o) files runs the risk of compatibility problems though.

If it is hard to get it to compile on Windows an alternative might also be putting it in a separate extension module that is not available on Windows. The same issue should come up for the `dstrf` module because it’s using the same c library, right?

proloyd:
Seems that building with a C++ compiler in Windows can fix the problem. Should we opt for that?
Will that be any problem?

christianbrodbeck:
If it works that seems fine, we can always revert if it causes a problem.

It should be made clear from the outside that the submodule is under a different license, maybe add a note to the `LICENSE` file? 

Also, since `dsyevc3` has itself no Python component, does it have to be inside a Python module, or could it be located outside in its own directory?

Are the changes to `setup.py` still needed?

proloyd:
Yeah. I just finalized `setup.py`. Hopefully it will pass the tests.
Secondly, yes the `dsyevc3.c` etc.  needs to be where it is, otherwise .pyx files won't be able to find them. (Right?) 
I don't get the `LICENSE` part? What do you mean by adding a note? I already included the particulat license notice in the `vector.pyx`. Do you mean like [this?](https://github.com/nipy/nibabel/blob/master/COPYING)

christianbrodbeck:
> Secondly, yes the dsyevc3.c etc. needs to be where it is, otherwise .pyx files won't be able to find them. (Right?)

Since `vector.pyx` contains the full path to the `dsyevc3` from module root I assumed it might be possible to place it outside ofthe Python module. This seemed like a cleaner separation between Python and included C code. If it’s not possible ignore my comment :)

> I don't get the LICENSE part? What do you mean by adding a note? I already included the particulat license notice in the vector.pyx. Do you mean like this?

Yes exactly, that would be great. Basically, someone coming to this repository should immediately see that they have to treat different components differently, without looking into sub folders.

proloyd:
The LICENSE file has been updated.

setup.py
```diff
@@ -49,17 +50,29 @@
 LooseVersion(version)  # check that it's a valid version
 
 # Cython extensions
-ext_paths = ('eelbrain/*%s', 'eelbrain/_trf/*%s', 'eelbrain/_stats/*%s')
+ext_c_paths = ('eelbrain/*%s', 'eelbrain/_trf/*%s', 'eelbrain/_stats/*%s')  # C
+ext_cpp_paths = ('eelbrain/_stats/*%s',)                                    # C++
```
christianbrodbeck: Won't this lead to potential conflict? It will find `eelbrain/_stats/*.pyx` twice. Might be simpler to just hard-code the extension paths, there are not that many...

setup.py
```diff
@@ -19,11 +19,12 @@
...
 
 from distutils.version import LooseVersion
-from distutils.extension import Extension
 from glob import glob
 from os.path import pathsep
 import re
-from setuptools import setup, find_packages
+from setuptools import setup, find_packages, Extension
+# To circumvent "error: each element of 'ext_modules' option must be an Extension instance or 2-tuple"
+# import Extension from setuptools instead of distutils.extension.
```
christianbrodbeck: No need for this comment

eelbrain/_stats/vector.pyx
```diff
@@ -1,5 +1,32 @@
...
+This library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public
+License along with this library; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301
+USA
+"""
```
christianbrodbeck, line 10: Actually, does it need that license statement here? There is no actual 3rd party code in this file right? It only links to it?

proloyd: Should I remove the license statement, and only state that this relies on the code available under GLPL? Or remove it totally?

christianbrodbeck: Maybe leave the comment that it relies on it.

proloyd: Cool.

christianbrodbeck:
Thanks @proloyd !!

PR #35, opened 2019-04-21
WIP: stc-to-dataset loader

reddigari:
Working back-end for the loader portion of the GUI we discussed. The tests don't use real stcs, but I've tested locally on one dataset and it works as expected.

Going to dive into the wx stuff this week. Planning to follow the patterns you use in the epoch- and component-selection GUIs, but any advice you might have would be awesome.


eelbrain/_io/tests/test_mrat.py
```diff
@@ -0,0 +1,48 @@
+import os
+import itertools
+from nose.tools import assert_in, assert_raises, assert_true, \
```
christianbrodbeck: I’m moving from `nose` to `pytest`, which gives better introspection when using simple `assert` statements in tests.

eelbrain/_io/mrat.py
```diff
@@ -0,0 +1,126 @@
...
+        if self.levels is None or self.factors is None:
+            return None
+        return " x ".join(map(str, self._level_lens))
+
+    def make_dataset(self, load_stcs=True):
+        """
+        Create a Dataset with one case per condition per subject, and source
+        estimates loaded as an NDVar.
+
+        Any `stc_kwargs` from __init__ are passed to `eelbrain.load.fiff.stc_ndvar()`.
```
christianbrodbeck: Code in `rst` uses two `

eelbrain/_io/mrat.py
```diff
@@ -0,0 +1,126 @@
...
+import re
+import glob
+import itertools
+from mne import read_source_estimate
+
+from .._data_obj import Dataset, Factor
+from .fiff import stc_ndvar
+
+
+class DatasetSTCLoader(object):
```
christianbrodbeck: `(object)` is redundant in Python 3

christianbrodbeck (approved):
I added some general comments, did you want to expose this loader in the API or only through the GUI?

For the GUI itself, the component/epoch selection GUIs are probably way more complicated than you need, you don’t need any embedded matplotlib plots, right? You probably just need a `wx.Dialog` with some controls. If you inherit from `EelbrainDialog` it will give basic integration. The `FindRareEventsDialog` here might be a useful example https://github.com/christianbrodbeck/Eelbrain/blob/master/eelbrain/_wxgui/select_components.py#L1117 If you let me know what functionality you would like I can point you towards relevant information. You’ve seen the wxPython demo with examples for all the different controls?

reddigari:
- Entirely up to you whether you expose it programmatically. I personally used to hack together this functionality when dealing with old data or helping out people who didn't use the experiment class. So I probably would have used it, but I don't know about others.
- For this GUI you're right that it'll be quite simple. When I get to the actual stats-running GUI we'll need  embedded matplotlib axes and a more dynamic interface. Is it possible to embed tvtk objects for interactively viewing clusters on brain surfaces?
- Am I correctly interpreting that the travis failure is from a docstring style issue and not a failed test?

christianbrodbeck:
> Entirely up to you whether you expose it programmatically. I personally used to hack together this functionality when dealing with old data or helping out people who didn't use the experiment class. So I probably would have used it, but I don't know about others.

Sounds like it would be useful for some, maybe check with the people who might most likely use it?

> For this GUI you're right that it'll be quite simple. When I get to the actual stats-running GUI we'll need embedded matplotlib axes and a more dynamic interface. Is it possible to embed tvtk objects for interactively viewing clusters on brain surfaces?

Yes, you can look at the functions in `plot.brain` for how to embed `pysurfer`/`mayavi` in wxPython. You can have multiple plots in one window through using `wx.Panel`s.

> Am I correctly interpreting that the travis failure is from a docstring style issue and not a failed test?

Yes, you can invoke that docstring check locally with `$ pydocstyle eelbrain`

christianbrodbeck:
This pull request **introduces 1 alert** when merging 2c91ad1c1a10334d3e000719c72c6cc398c586f3 into f953933be140fd3a2ed86a656ea33d4c2c7e6058 - [view on LGTM.com](https://lgtm.com/projects/g/christianbrodbeck/Eelbrain/rev/pr-7a81afc4afaa5936e609123fb0a8998d93e07bcb)

**new alerts:**

* 1 for Unused local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

reddigari:
How long do the GUI tests typically take to run? Started to write mine (which pass very quickly), but it hangs for at least 5 minutes on `test_select_components.py` and my computer gets all hot and bothered. I'm running locally with `pythonw $(which pytest) eelbrain/_wxgui`.

christianbrodbeck:
This pull request **introduces 1 alert** when merging 76061dde9ab8e6fc6119d6b1c7063ea89234d8c8 into f953933be140fd3a2ed86a656ea33d4c2c7e6058 - [view on LGTM.com](https://lgtm.com/projects/g/christianbrodbeck/Eelbrain/rev/pr-1ea05704640c27eee1cac16eefd730c55e4e9ed7)

**new alerts:**

* 1 for Unused local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

christianbrodbeck:
`$ pytest eelbrain/_wxgui` passes in about 10 s for me (both in master and for this PR). I also just recreated the environment to make sure it's not due to some package that changed, it still took < 10 s. Can you run the commands of the tests individually in a terminal to figure out where it hangs?

(BTW you can also patch `pytest` with `$ ./fix-bin pytest` so you don't need to use `$(which pytest)`)

reddigari:
Ahh it was downloading the MNE test data. All good now, thanks

christianbrodbeck:
This pull request **introduces 1 alert** when merging 69e482c15f12430d1eee0adce5eb3302470e6491 into f953933be140fd3a2ed86a656ea33d4c2c7e6058 - [view on LGTM.com](https://lgtm.com/projects/g/christianbrodbeck/Eelbrain/rev/pr-4e4ed4753a5d5f56b5dc628c6eb52e2a977a5bbd)

**new alerts:**

* 1 for Unused local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

reddigari:
Trying to add a second test, but it seems the presence of a second test method (specifically the presence of `gui.load_stcs()` results in a seg fault. The odd part is that the test passes before the fault. Any ideas? Neither of the other GUI tests has more than one method.
Selection_646

christianbrodbeck:
This pull request **introduces 1 alert** when merging 697abe23297655ccd158a6e17b12398b89a619a1 into 7190ec7b43147fe7aa1d68048c1afbdb76b4e190 - [view on LGTM.com](https://lgtm.com/projects/g/christianbrodbeck/Eelbrain/rev/pr-26f99827b37c9aaf04d36e5856c3cc7072e8a4b4)

**new alerts:**

* 1 for Except block handles &#39;BaseException&#39;

---

*Comment posted by [LGTM.com](https://lgtm.com)*

christianbrodbeck:
If I simply duplicate the existing test (and rename the copy) I don't get a segmentation fault. Looks like it might be platform-specific... Can you get a segmentation fault by executing the relevant commands from the iPython terminal?

```
(eeldev) Christians-Mac-mini ~/Code/Eelbrain$ pytest eelbrain/_wxgui/tests/test_load_stcs.py 
============================================ test session starts =============================================
platform darwin -- Python 3.7.0, pytest-4.4.1, py-1.8.0, pluggy-0.9.0
rootdir: /Users/christian/Code/Eelbrain, inifile: setup.cfg
plugins: faulthandler-1.5.0, cov-2.6.1
collected 2 items                                                                                            

eelbrain/_wxgui/tests/test_load_stcs.py ..                                                             [100%]

========================================== 2 passed in 4.22 seconds ==========================================
```

reddigari:
Hmmm yeah works fine on macOS for me too. Let's see if it's replicated on Travis.
EDIT: yep, segfault on the Linux build, fine on macOS

At least one person has said they would use the loader from the API. Are you okay exposing the class in `eelbrain.load`? I'm thinking that "mrat" is not particularly meaningful anymore, unless you want to name this whole feature mrat to be consistent with the output methods on the experiment class.

I have started on the stats GUI, but I think I'd like to continue it on a new feature branch. Do you have a preference on merging this now vs. when the whole thing is more fleshed out?

christianbrodbeck:
> yep, segfault on the Linux build, fine on macOS

Strange, if it's just due to having two test functions in one file maybe just put the second test in a separate file? If you want to investigate I would check the `wx`/`pytest` issues, and if there is nothing like it try to come up with a minimal example, but it might be a rabbit hole :) 

> At least one person has said they would use the loader from the API. Are you okay exposing the class in eelbrain.load? 

Yes of course

> I'm thinking that "mrat" is not particularly meaningful anymore, unless you want to name this whole feature mrat to be consistent with the output methods on the experiment class.

I agree. Is the MneExperiment `mrat` export function still used anyways? I've been tempted to remove it but since it does not need any maintenance I've left it so far.

Maybe you could expose the loader in `load.fiff`, next to `stc_ndvar`, as `stc_dataset` or something like that? Since basically instead of loading a single `stc` it loads multiple `stcs`, right? (Really `fiff` is a slight misnomer here since `stcs` are not stored in `fiff` format, it should probably be `load.mne`, but that's a different issue... although if you want to make the API more consistent I am not opposed :)

> I have started on the stats GUI, but I think I'd like to continue it on a new feature branch. Do you have a preference on merging this now vs. when the whole thing is more fleshed out?

I think in general it is preferable to merge in smaller chunks once they are finished, makes it less likely for two branches to become so different that merge conflicts arise

reddigari:
Explicitly closing the Frame at the end of each test took care of it.

> Maybe you could expose the loader in load.fiff, next to stc_ndvar, as stc_dataset or something like that?

That's what I did for now, `load.fiff.stc_dataset` takes a path and returns a loader object. I documented the use with an example in the docstring.

> but that's a different issue... although if you want to make the API more consistent I am not opposed :)

Ooooh look who's not afraid of breaking changes; do I smell a 1.0.0 release?  😄 

If all the builds pass, I think this is good to go. Let me know if I'm missing anything!

christianbrodbeck:
This pull request **introduces 1 alert** when merging 18d34de200a4b294b720e2a9ac516ac10e1d48d6 into 62b364e1da9f73aa7104565b1adbef7fa3386788 - [view on LGTM.com](https://lgtm.com/projects/g/christianbrodbeck/Eelbrain/rev/pr-8fa9d14c9bc5796e84338a72b31805cb9be8ee04)

**new alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

eelbrain/load/fiff.py
```diff
@@ -94,3 +94,5 @@
     forward_operator, inverse_operator, mne_epochs, mne_raw, raw_ndvar,
     sensor_dim, stc_ndvar
 )
+
+from .._io.stc_dataset import stc_dataset
```
christianbrodbeck: There’s supposed to be a newline at the end of the file, I did not know why, but stack overflow has an answer :) https://stackoverflow.com/a/729795/166700

eelbrain/gui.py
```diff
@@ -168,3 +168,22 @@ def select_epochs(
...
         run()
+
+def load_stcs():
+    """GUI for detecting and loading source estimates
+
+    Notes
+    -----
+    This GUI prompts the user for the directory containing the
+    source estimates, then automatically detects the experimental
+    design, and loads the data into an ``eelbrain.Dataset``.
```
christianbrodbeck:
```suggestion
    design, and loads the data into a :class:`Dataset`.
```

Sphinx will add a hyperlink when marked this way

eelbrain/gui.py
```diff
@@ -168,3 +168,22 @@ def select_epochs(
         return frame
     else:
         run()
+
+def load_stcs():
```
christianbrodbeck, line 173: Shouldn't this function return the `Dataset` with the STCs?

eelbrain/_io/stc_dataset.py
```diff
@@ -0,0 +1,167 @@
...
+        return des
+
+    def make_dataset(self, load_stcs=True, subject="fsaverage",
+                     src="ico-4", **stc_kwargs):
+        """
+        Load stcs into a Dataset with columns for subject and experimental factors
+
+        Dataset contains one case per condition per subject, and source estimates
+        loaded as an NDVar. Any additional keyword arguments are passed to
+        ``eelbrain.load.fiff.stc_ndvar()``. If SUBJECTS_DIR is not set int your
```
christianbrodbeck:
```suggestion
        :meth:`eelbrain.load.fiff.stc_ndvar`. If ``SUBJECTS_DIR`` is not set in your
```

That should add a hyperlink to the function

eelbrain/_io/stc_dataset.py
```diff
@@ -0,0 +1,167 @@
...
+    ----------
+    path : str
+        Data directory passed to :class:`DatasetSTCLoader` constructor
+
+    Returns
+    -------
+    loader : :class:`DatasetSTCLoader`
+        Loader object with levels detected
+    """
+    return DatasetSTCLoader(path)
```
christianbrodbeck: If a `DatasetSTCLoader` is returned, the class should probably also be in the documentation (the reference page). Do you need the `stc_dataset` function or could you directly expose the `DatasetSTCLoader`? Alternatively, could the parameters (only `factor_names`?) be exposed in the function so that the class can remain private?

christianbrodbeck (commented):
Looks great! Just a couple of small comments and questions.

I noticed one cosmetic issue: when I set an invalid directory, and then change to a good one, the error message in the status bar does not go away.

Can you also add links to `doc/reference.rst`?

eelbrain/gui.py
christianbrodbeck: Or, alternatively, it could ask for a name and place the dataset in the namespace (like `App.Attach()`, https://github.com/christianbrodbeck/Eelbrain/blob/master/eelbrain/_wxgui/app.py#L385), then you would not need to block the Terminal.

eelbrain/_io/stc_dataset.py
reddigari: I think it's best to expose the loader directly, as I have a feeling that fully specifying experiment structure will in practice require more methods than `set_factor_names`. Should I just make it `load.fiff.DatasetSTCLoader`?

eelbrain/gui.py
reddigari: This was intended as the command to launch the loader GUI. I went with `load_stcs` to be consistent with `select_epochs` and `select_components`, but I'm happy to change it to whatever you suggest

reddigari:
Fixed the GUI bug (good catch, thanks) and addressed the style/docstring issues. Let me know what you think about the API decisions, and I'll update `reference.rst`

christianbrodbeck:
This pull request **introduces 1 alert** when merging 3e9cc3dc9decd2905f1e236a16e8942a42d7b61d into 62b364e1da9f73aa7104565b1adbef7fa3386788 - [view on LGTM.com](https://lgtm.com/projects/g/christianbrodbeck/Eelbrain/rev/pr-c945924825b5fc240c125327978e6716d416e918)

**new alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

eelbrain/_io/stc_dataset.py
christianbrodbeck: Yes, sounds good

eelbrain/gui.py
christianbrodbeck: > This was intended as the command to launch the loader GUI. 

Right, I got that, but would you not want the `Dataset` returned from this command? I.e. use the GUI to organize your STCs and then have access to them in a `Dataset` object, kind of like `load.unpickle()` when called without argument? Or was the intention for this to launch the stats GUI?

reddigari: Oooooh I see, I hadn't thought of that. I envisioned it opening the stats GUI.

christianbrodbeck:
This pull request **introduces 1 alert** when merging ce3012d95f89281006117e6b6f8dd69e36cb23ff into 62b364e1da9f73aa7104565b1adbef7fa3386788 - [view on LGTM.com](https://lgtm.com/projects/g/christianbrodbeck/Eelbrain/rev/pr-0fc082c3e6b933503ca7d1cf30c893fbfc6868c2)

**new alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

eelbrain/load/fiff.py
```diff
@@ -88,9 +89,27 @@
 
     >>> ds = load.fiff.events(raw_file_path, events=events_file_path)
 
+
+Loading source estimates into a Dataset
+---------------------------------------
+
+Previously exported stc files can be loaded into a :class:Dataset with the
```
christianbrodbeck:
```suggestion
Previously exported stc files can be loaded into a :class:`Dataset` with the
```

eelbrain/load/fiff.py
```diff
@@ -88,9 +89,27 @@
...
+:class:`DatasetSTCLoader` class. The stcs must reside in subdirectories named
```
christianbrodbeck: I think this link might fail to generate a link because it's expecting objects from the `eelbrain` namespace (the `.. currentmodule:: eelbrain` line above; not sure though whether that means it will not inspect the local namespace of the docstring). You'd need to specify the path relative to `eelbrain` (```:class:`load.fiff.DatasetSTCLoader` ```). If you still want to show it as `DatasetSTCLoader` in the docs for easier reading, you can spell it ```:class:`~load.fiff.DatasetSTCLoader` ```.

You can build the docs with `$ make doc` to test the that the links work (unfortunately `readthedocs` don't support building PRs yet). I got some issue with the latest sphinx version, so if your env has the latest sphinx version you might have to rebase onto master to build them.

eelbrain/gui.py
```diff
@@ -168,3 +168,23 @@ def select_epochs(
...
+    """
+    from ._wxgui.app import get_app
+    from ._wxgui.load_stcs import STCLoaderFrame, TEST_MODE
+
+    get_app()
+    frame = STCLoaderFrame(None)
+    if TEST_MODE:
+        return frame
+    else:
+        run()
```
christianbrodbeck: This is missing a newline

reddigari:
Links are fixed (docs built okay). Thanks for bearing with me, I'll browse through the rst and Sphinx reference a little more closely this week.

eelbrain/gui.py
christianbrodbeck: Did you want to add the functionality to return the `Dataset`? Otherwise I would add it with the PR that adds the actual GUI (and maybe call is something like `stats_viewer` rather than `load_stc`)?

christianbrodbeck: Actually, if exposing the loader to return a `Dataset`, it might more naturally go into `load.fiff`, analogous to functions like `load.unpickle`, which also open a file dialog if called without a filename.

reddigari: So you mean load.fiff.stc_dataset() would itself launch the GUI and return the dataset?

 If so, how about a parameter in `gui.load_stcs()` (e.g. `launch_stats=False`) that decides whether the submit button returns the dataset or passes it on to the stats GUI? (For now it could raise `NotImplementedError` or something until the stats GUI is ready.)

christianbrodbeck:
Sounds good!

eelbrain/gui.py
christianbrodbeck: Sorry for the delay,

> Sounds good

Was intended as reply to this post :) Email reply seems to have added it to the bottom of the PR.

> So you mean load.fiff.stc_dataset() would itself launch the GUI and return the dataset?

Either two separate functions (e.g. `gui.load_stcs` and `gui.stc_stats`) or a parameter to `load_stcs` both seem like good options to me.

reddigari: Am I understanding correctly that I need to call `run(block=True)`, make the loaded dataset an attribute of the frame while the main loop is running, and then return`frame.ds` once the user kills the GUI? If you prefer the `App.Attach()`route I'll go with that, but I'm unfamiliar with it.

reddigari: Actually, after playing around with it, I'm not a fan of blocking the terminal.  Regardless of how we launch the stats GUI, it will always require the user to kill the loader GUI so that the dataset can be retrieved and passed on (or returned).

I think the right approach is a `launch_stats` parameter in `gui.load_stcs`, which is passed into the constructor of `STCLoaderFrame`. In the `OnSubmit` event, if `launch_stats` is true, the frame instantiates a stats GUI frame and directly passes on the dataset. If it's false, it calls `App.Attach()` to give the dataset back to the terminal. No blocking needed.

christianbrodbeck: What about having two buttons at the bottom of `STCLoaderFrame`, one to place the dataset in the gloabel namespace ("attach" it) and a second one to launch the stats GUI? That would reduce the amount of typing in arguments?

reddigari: Works for me. Should `load.fiff.stc_dataset()` still open the GUI, or are we ditching that?

christianbrodbeck: You mean it would just be an alias for the same function? If so I would vote against that, to keep the API simpler.

reddigari:
I've left `DatasetSTCLoader` in the `load.fiff` module and its docs; let me know if you want it removed or moved elsewhere

eelbrain/testing/_testing.py
```diff
@@ -115,6 +115,7 @@ class GUITestContext(ContextDecorator):
         eelbrain._wxgui.select_epochs,
         eelbrain._wxgui.select_components,
         eelbrain._wxgui.history,
+        eelbrain._wxgui.load_stcs
```
christianbrodbeck: Just nitpicking :) but if you add a comma at the end of this line then the next addition would not need to modify this line.
```suggestion
        eelbrain._wxgui.load_stcs,
```

eelbrain/_io/stc_dataset.py
```diff
@@ -0,0 +1,155 @@
...
+    Notes
+    -----
+    When instantiated, the loader will automatically do level detection
+    based on .stc filenames. The user must explicitly set the factor
+    names with :meth:`DatasetSTCLoader.set_factor_names`. The dataset
+    may then be loaded via :meth:`DatasetSTCLoader.make_dataset`.
+
+    Examples
+    --------
+    >>> loader = DatasetSTCLoader("path/to/exported/stcs")
```
christianbrodbeck, line 41: Does this render correctly? Sometimes an indent is needed for code, but I am not sure if that applies if it's the first lime after the heading?

eelbrain/_io/stc_dataset.py
```diff
@@ -0,0 +1,155 @@
...
+    >>> loader.set_factor_names(["factor1", "factor2"])
+    >>> ds = loader.make_dataset(subjects_dir="mri/")
+    """
```
christianbrodbeck, line 48: You could add cross-references between `DatasetSTCLoader` and `gui.load_stcs` using the `See Also` section

eelbrain/_io/stc_dataset.py
```diff
@@ -0,0 +1,155 @@
+import os
+import re
+import glob
+import itertools
+from mne import read_source_estimate
```
christianbrodbeck, line 6:
```suggestion

from mne import read_source_estimate
```
Separate standard library and 3rd party imports

eelbrain/_io/stc_dataset.py
```diff
@@ -0,0 +1,155 @@
...
+        self.levels = None
+        self.factors = None
+        self._n_factors = None
+        self._level_lens = None
+        self._find_subjects()
+        self._find_level_names()
+
+    def __repr__(self):
+        tmp = "<DatasetSTCLoader: {} subjects | {} design>"
+        return tmp.format(len(self.subjects), self.design_shape)
```
christianbrodbeck, line 63: Just to make you aware of the new formatted string literals (https://docs.python.org/3/whatsnew/3.6.html#pep-498-formatted-string-literals) which could make this as simple as

```python
return f"<DatasetSTCLoader: {len(self.subjects)} subjects | {self.design_shape} design>"
```

christianbrodbeck (commented):
Looks good, I've left a few minor comments

> I've left `DatasetSTCLoader` in the load.fiff module and its docs; let me know if you want it removed or moved elsewhere

That seems like the right place for it.

Let me know when it's ready for merging!

eelbrain/_io/stc_dataset.py
reddigari: It does

eelbrain/_io/stc_dataset.py
reddigari: I tend to avoid them for the sake of 2/3 compatibility and line length, but I should probably get in the habit

christianbrodbeck: Ah okay with me, I just wanted to mention it because its a new feature I find really convenient

eelbrain/_io/stc_dataset.py
reddigari: Done

eelbrain/_io/stc_dataset.py
```diff
@@ -0,0 +1,161 @@
...
+    --------
+    >>> loader = DatasetSTCLoader("path/to/exported/stcs")
+    >>> loader.set_factor_names(["factor1", "factor2"])
+    >>> ds = loader.make_dataset(subjects_dir="mri/")
+
+    See Also
+    --------
+    :func:`eelbrain.gui.load_stcs` : a GUI to load source estimates
+     into a Dataset
+    """
```
christianbrodbeck, line 48: Do these render correctly? You don't need `:func:`, and I thought they ought to be one cross-ref per line (but I might be wrong about that; https://numpydoc.readthedocs.io/en/latest/format.html)

reddigari: They render like the other `See Also` examples in the docs, but you're right about the prefixes. Removed (I amended the commit and force pushed, so we should be good to go)

Screen Shot 2019-05-18 at 3 05 43 PM

christianbrodbeck: Great, thanks!

christianbrodbeck:
Thanks @reddigari !

PR #39, opened 2020-02-27
Trf len

jpkulasingham:


eelbrain/_experiment/mne_experiment.py
```diff
@@ -6348,7 +6348,10 @@ def _set_analysis_options(self, data, baseline, src_baseline, pmin, tstart, tsto
 
         # time window
         if tstart is not None or tstop is not None:
-            items.append(_time_window_str((tstart, tstop)))
+            if isinstance(tstart, (tuple, list, np.ndarray)):
+                tstart1 = min(tstart)
+                tstop1 = max(tstop)
+            items.append(_time_window_str((tstart1, tstop1)))
```
christianbrodbeck: Can you drop this for now (for this PR)? I think we'll have to handle this primarily through the `TRFExperiment` class.

eelbrain/_trf/_boosting.py
```diff
@@ -399,32 +398,63 @@ def boosting(y, x, tstart, tstop, scale_data=True, delta=0.005, mindelta=None,
...
+            raise ValueError(
+                f'mismatched len(tstart) = {len(tstart)}, len(tstop) = {len(tstop)}')
+        elif len(tstart) == 0:
+            raise ValueError(f'tstart cannot be an empty list')
+        if len(tstart) != n_x:
+            if len(tstart) == 1:
+                tstart = [tstart[0] for _ in range(n_x)]
+                tstop = [tstop[0] for _ in range(n_x)]
+            elif len(tstart) == len(data.x_name):
+                print(f'Warning: len(tstart) matches len(x) = {len(x)} but not # of predictor dimensions = {n_x}. Infering from data._x_meta')
```
christianbrodbeck: I would not warn every time, this is a valid way of specifying `tstart`

eelbrain/_trf/_boosting.py
```diff
@@ -587,9 +619,22 @@ def boost(y, x, x_pads, all_index, train_index, test_index, i_start, trf_length,
     error = ERROR_FUNC[error]
     n_stims, n_times = x.shape
     assert y.shape == (n_times,)
-
-    h = np.zeros((n_stims, trf_length))
-
+    if isinstance(i_start, (tuple, list, np.ndarray)):
```
christianbrodbeck: Doesn't the outer function make sure that `i_start` is always an integer array? Generally, it's a good idea to make sure the data conform to a specific type in the outermost function, then you can avoid making multiple redundant checks inside of loops.

eelbrain/_trf/_boosting.py
```diff
@@ -519,12 +549,12 @@ def boosting(y, x, tstart, tstop, scale_data=True, delta=0.005, mindelta=None,
...
         debug_attrs = {}
 
-    h = data.package_kernel(h_x, tstart)
+    h = data.package_kernel(h_x, min(tstart))
     model_repr = None if model is None else data.model
     prefit_repr = None if prefit is None else repr(prefit)
     return BoostingResult(
         # input parameters
-        data.y_name, data.x_name, tstart, tstop, scale_data, delta, mindelta, error,
+        data.y_name, data.x_name, min(tstart), max(tstop), scale_data, delta, mindelta, error,
```
christianbrodbeck: It would probably be useful for the result to preserve the input argument (whether a scalar or a sequence) so you can later double check what you used for fitting the model – the min/max could be taken internally in the result object where necessary.

eelbrain/_trf/_boosting.py
```diff
@@ -658,9 +703,15 @@ def boost(y, x, x_pads, all_index, train_index, test_index, i_start, trf_length,
                 break
 
         # generate possible movements -> training error
-        generate_options(y_error, x, x_pads, x_active, train_index, i_start, delta_error_func, delta, new_error, new_sign)
-
-        i_stim, i_time = np.unravel_index(np.argmin(new_error), h.shape)
+        generate_options(y_error, x, x_pads, x_active, train_index, i_start, delta_error_func, delta, new_error,
+                         trf_length, new_sign)
+        min_err = 99999999
+        for i_st in range(n_stims):
```
christianbrodbeck: `for` loops are slow – could we instead just set `new_error` to a large number on unused indices? `np.inf`?

eelbrain/_trf/_boosting.py
```diff
@@ -691,6 +742,13 @@ def boost(y, x, x_pads, all_index, train_index, test_index, i_start, trf_length,
         for step in history[-1: best_iteration: -1]:
             if step.delta:
                 h[step.i_stim, step.i_time] -= step.delta
+        h_len = max(i_stop) - min(i_start)
+        h_i_start = i_start - min(i_start)
+        h_i_stop = [h1 + h2 for h1, h2 in zip(h_i_start, trf_length)]
+        h_new = np.zeros((n_stims, h_len))
+        for i_stim in range(n_stims):
+            h_new[i_stim, h_i_start[i_stim]:h_i_stop[i_stim]] = h[i_stim, :trf_length[i_stim]]
+        h = h_new
```
christianbrodbeck: Woulnd't it be more efficient to represent `h` like this from the beginning – with time aligned between predictors and zeros at disabled elements – then a lot of other functions, like `convolve`, would not have to be modified at all?

env-dev.yml
```diff
@@ -1,7 +1,7 @@
 # Environment for development
 # usage: $ conda env create --file=env-dev.yml
 # to run stats test, install r-ez with $ conda install -c ncil r-ez --no-deps
-name: eeldev
+name: eelfork
```
christianbrodbeck: This should not be in the PR :) you can also specify the name when creating the environment list

`$ conda env create --file env.yml -n eelfork`

christianbrodbeck (changes requested):
Thanks! I've added some initial comments and questions, let me know if any of them are not clear.

eelbrain/_experiment/mne_experiment.py
jpkulasingham: Ok. Will remove

eelbrain/_trf/_boosting.py
jpkulasingham: Removed

eelbrain/_trf/_boosting.py
jpkulasingham: I can remove this check, but `test_boosting_func()` calls `boost()` directly, so the test doesn't pass. Should I modify `test_boosting_func()` to give in a nparray tstart tstop?

eelbrain/_trf/_boosting.py
jpkulasingham: Ok will add

eelbrain/_trf/_boosting.py
jpkulasingham: Will change

eelbrain/_trf/_boosting.py
jpkulasingham: Yes, but then several other parts of the function have to be changed in order to align for updating the error. And `convolve` may have to be modified anyway since the padding will differ based on where the convolution should start and stop. Is this essential?

env-dev.yml
jpkulasingham: Removed

christianbrodbeck: Did you forget to push the modification?

eelbrain/_trf/_boosting.py
christianbrodbeck: Yes that's okay, since `boost` is not a public function it's okay to change its API (and corresponding tests)

eelbrain/_trf/_boosting.py
```diff
@@ -193,8 +192,9 @@ def __repr__(self):
             x = ' + '.join(map(str, self.x))
         items = [
             'boosting %s ~ %s' % (self.y, x),
-            '%g - %g' % (self.tstart, self.tstop),
         ]
+        for t1, t2 in zip(self.tstart, self.tstop):
+            items.append('%g - %g' % (t1, t2))
```
christianbrodbeck: The boosting result has to be able to handle `tstart` and `tstop` as scalars, since they will be scalars in older, pickled results. I also think that it would be much more readable to only display one range if a single range was used. If multiple ranges were used, might it be more readable to interleave them with the predictors? i.e., 

`<boosting y ~ word (0 - 0.400) + sentence (0 - 1.300)>`
vs
`<boosting y ~ word + sentence, 0 - 0.400, 0 - 1.300>`

eelbrain/_trf/tests/test_boosting.py
```diff
@@ -224,3 +225,56 @@ def test_boosting_func():
     assert rr == approx(mat['crlt'][1, 0])
     # svdboostV4pred multiplies error by number of predictors
     assert_allclose(test_sse_history, mat['Str_testE'][0] / 3)
+
+@pytest.mark.parametrize('n_workers', [0, True])
+def test_trf_len(n_workers):
+    configure(n_workers=n_workers)
+    # test vanilla boosting
+    x = NDVar(np.random.normal(0, 1, 1000), UTS(0, 0.1, 1000),name='x')
```
christianbrodbeck: I would use a seeded random state to avoid random surprises in test runs
```suggestion
    rng = np.random.RandomState(0)
    x = NDVar(rng.normal(0, 1, 1000), UTS(0, 0.1, 1000), name='x')
```
...

eelbrain/_trf/_boosting.py
christianbrodbeck: I'll have to think about it when I look into the code in details, leave that for now

christianbrodbeck (changes requested):
Could you change the way the result is initialized such that `result.tstart` and `.tstop` are scalars if the user-provided argument is scalar? Turning them into arrays breaks a couple of things that currently work, like comparison `result.tstart == 0`.

Let me know when that API issue is done, then I'll have a closer look at the implementation issue.

eelbrain/_trf/_boosting.py
jpkulasingham: Ok changed `test_boosting_func()`

eelbrain/_trf/_boosting.py
jpkulasingham: changed accordingly

eelbrain/_trf/tests/test_boosting.py
jpkulasingham: added seed

eelbrain/_trf/_boosting.py
christianbrodbeck: I would keep the user input and variable you use internally separate. For example, the current 
```
    if len(tstart) == 1:
        tstart = tstart[0]
```
would get it wrong if the user used `tstart=[0]`, or if the user used `tstart=0` and there are more than one `x`.

eelbrain/_trf/tests/test_boosting.py
```diff
@@ -212,9 +215,10 @@ def test_boosting_func():
     y = mat['signal'][0]
     x = mat['stim']
     x_pads = np.zeros(len(x))
-
+    tstart = np.asarray([0, 0, 0])
+    tstop = np.asarray([10, 10, 10])
```
christianbrodbeck:
```suggestion
    tstart = np.array([0, 0, 0], np.int64)
    tstop = np.array([10, 10, 10], np.int64)
```

eelbrain/_trf/tests/test_boosting.py
```diff
@@ -195,8 +196,10 @@ def test_boosting_func():
     all_segments = np.array([[0, seg_len], [seg_len, y_len]], np.int64)
     train_segments = all_segments[1:]
     test_segments = all_segments[:1]
+    tstart = np.asarray([0])
+    tstop = np.asarray([10])
```
christianbrodbeck:
```suggestion
    tstart = np.array([0], np.int64)
    tstop = np.array([10], np.int64)
```

christianbrodbeck: (On a 32 bit OS, using integers defaults to 32 bit integers (long)...)

eelbrain/_trf/tests/test_boosting.py
christianbrodbeck: I think you'll need to commit this one too, no?

christianbrodbeck (approved):

christianbrodbeck:
Great, ready to merge from your end?

jpkulasingham:
> Great, ready to merge from your end?

Yup. Fine with me :)

christianbrodbeck:
Thanks @jpkulasingham !

PR #49, opened 2022-07-08
FIX read_cnd extChan

jpkulasingham:
read_cnd used to load eeg data into extChan as well. Fixed it to load extChan data.


christianbrodbeck:
Thanks @jpkulasingham

PR #80, opened 2025-04-07
Fixed nilearn dependencies and removed cbar_vmin/vmax redundancies

huzafahyde:
directly imported nilearn functions, and updated the _crop_colorbar function to the newest version. Removed redundant cbar_vmin and cbar_vmax

### DIFF
--- a/eelbrain/plot/_glassbrain.py
+++ b/eelbrain/plot/_glassbrain.py
@@ -46,6 +46,7 @@ from .._utils.numpy_utils import newaxis
 from ._base import ColorBarMixin, TimeSlicerEF, Layout, EelFigure, brain_data, butterfly_data, use_inline_backend
 from ._utsnd import Butterfly
 
+import numbers
 
 # Copied from nilearn.image.resampling to avoid sklearn import with forced warnings
 def get_bounds(shape, affine):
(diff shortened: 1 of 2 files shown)


eelbrain/plot/_glassbrain.py
```diff
@@ -46,6 +46,7 @@
 from ._base import ColorBarMixin, TimeSlicerEF, Layout, EelFigure, brain_data, butterfly_data, use_inline_backend
 from ._utsnd import Butterfly
 
+import numbers
```
christianbrodbeck: Can you move this up into the first block of imports? (The order of the blocks is standard library, then other libraries, then imports from the current library)

eelbrain/plot/_glassbrain.py
```diff
@@ -155,10 +198,10 @@ class GlassBrain(TimeSlicerEF, ColorBarMixin, EelFigure):
         Naturally, for this to work ``ndvar`` needs to contain space dimension
         (i.e 3D vectors). By default it is set to ``True``.
     symmetric_cbar
-        Specifies whether the colorbar should range from -vmax to vmax
-        or from vmin to vmax. Setting to 'auto' will select the latter if
+        Specifies whether the colorbar should range from -vmax to vmax (True)
+        or from vmin to vmax (False). Setting to 'auto' will select the latter if
         the range of the whole image is either positive or negative.
-        Note: The colormap will always be set to range from -vmax to vmax.
+        # Note: The colormap will always be set to range from -vmax to vmax.
```
christianbrodbeck: Did you mean to remove this line?

eelbrain/plot/_glassbrain.py
```diff
@@ -375,6 +415,63 @@ def __init__(
         TimeSlicerEF.__init__(self, 'time', time, display_text=time_text)
         self._show()
 
+    # Copied from nilearn.plotting.img_plotting
+    def _get_colorbar_and_data_ranges(self,
```
christianbrodbeck: If this does not use `self`, could this be included as a function (like the other functions you copied), instead of as a method?

eelbrain/plot/_glassbrain.py
```diff
@@ -375,6 +415,63 @@ def __init__(
         TimeSlicerEF.__init__(self, 'time', time, display_text=time_text)
         self._show()
 
+    # Copied from nilearn.plotting.img_plotting
```
christianbrodbeck: If you actually changed it (removing `cbar_vmin` if I understood that correctly) you could say "adapted from"

christianbrodbeck (changes requested):
Great start. I've added some stylistic comments in the code. Let me know if anything is unclear!

I have a request for the functionality: can we make the defaults more intuitive? I've tried using different arguments in the `volume-source-space` example script. For example, using no parameters at all (`plot.GlassBrain(y, title=f"{side} ear tone")`) results in the plot below, using a symmetric color-map but with vmin=0. To be consistent with the rest of the library, the default should probably use `SYMMETRIC_CMAPS` to determine whether `vmin` should be `-vmax` or `0`.

download

huzafahyde:
Addressed all review comments. For the defaults I have the logic set as such:
If there is no cmap specified by the user, and symmetric_cbar = 'auto' (ie if the user does not specify input parameters), it will set it to True or False depending on the datatype (False if bool, True if not) (lines 384-390). If the user wants to force symmetric_cbar, it will accept that input instead.

eelbrain/plot/_glassbrain.py
```diff
@@ -65,27 +65,126 @@ def get_bounds(shape, affine):
     box = np.dot(affine, box)[:3]
     return list(zip(box.min(axis=-1), box.max(axis=-1)))
 
+# Copied from nilearn.plotting.displays._slicers
```
christianbrodbeck, line 71: Could you include two blank lines before function/class definitions (generally we are trying to follow https://peps.python.org/pep-0008/, except for line length, because that often results in somewhat awkward code in my opinion)

christianbrodbeck (commented):
It's now fixed for the default `cmap`, but when I specify another one I get `vmin=0` – is there a reason not to use `SYMMETRIC_CMAPS` as default?
Screenshot 2025-04-09 at 1 55 28 PM

eelbrain/plot/tests/test_glassbrain.py
```diff
@@ -20,6 +20,30 @@ def test_glassbrain():
     p = plot.GlassBrain(ndvar_30, draw_arrows=False)
     p.close()
 
+    # vmin, vmax, and symmetric colorbar
+    p = plot.GlassBrain(ndvar_30, colorbar = True, cmap = 'lux-a')
```
christianbrodbeck:
```suggestion
    p = plot.GlassBrain(ndvar_30, colorbar=True, cmap='lux-a')
```
Small stylistic change, also below (https://peps.python.org/pep-0008/#other-recommendations)

christianbrodbeck:
Looks great now, could you just address the small PEP8 formatting issue? The test error is because the download for a test dataset was failing, I think it's working more reliably again now.

christianbrodbeck:
Test failure is unrelated. Thanks @huzafahyde !

PR #86, opened 2025-05-17
MAINT add pyproject.toml

proloyd:
Add `pyproject.toml` to modernize build process.

NOTE: setup.py is not defunct, in fact, it does the heavy lifting for creating the C-extensions as before. Only the ornamental information and build/installation requirements  were moved to `pyproject.toml`.

### DIFF
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -13,24 +13,30 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python 3.10
-      uses: actions/setup-python@v3
-      with:
-        python-version: '3.10'
+    # - name: Set up Python 3.10
+    #   uses: actions/setup-python@v3
+    #   with:
+    #     python-version: '3.10'
     - if: matrix.os == 'ubuntu-latest'
       name: Install dependencies
       run: |
         sudo apt-get update
-    - name: Install Mamba
+    - name: Use Mamba to create env
       uses: mamba-org/setup-micromamba@v2
       with:
-        environment-file: env-dev.yml
         environment-name: eeldev
+        create-args: >-
+          python=3.10
+          libcblas
+          numpy>=1.20
+          wxpython>=4.1.0
+          rpy2
+          r-car
         init-shell: bash
         cache-environment: true
         post-cleanup: 'all'
-    - name: build Eelbrain
-      run: pip install -e .
+    - name: Build eelbrain
+      run: pip install -e .[all]
       shell: bash -el {0}
     - if: matrix.os == 'ubuntu-latest'
       name: Test without GUI
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -13,10 +13,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    # - name: Set up Python 3.10
-    #   uses: actions/setup-python@v3
-    #   with:
-    #     python-version: '3.10'
     - if: matrix.os == 'ubuntu-latest'
       name: Install dependencies
       run: |
@@ -24,14 +20,8 @@ jobs:
     - name: Use Mamba to create env
       uses: mamba-org/setup-micromamba@v2
       with:
+        environment-file: env-dev.yml
         environment-name: eeldev
-        create-args: >-
-          python=3.10
-          libcblas
-          numpy>=1.20
-          wxpython>=4.1.0
-          rpy2
-          r-car
         init-shell: bash
         cache-environment: true
         post-cleanup: 'all'
--- a/eelbrain/datasets/_alice.py
+++ b/eelbrain/datasets/_alice.py
@@ -1,9 +1,7 @@
 # Helper to download data for Alice example
-import os
-import pooch
-
 from pathlib import Path
-import zipfile
+
+import pooch
 
 from .._types import PathArg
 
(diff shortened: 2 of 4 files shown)


proloyd:
@christianbrodbeck looks like `urllib.request.urlretreive()` is the culprit.

christianbrodbeck:
Yes, it's the download of the Alice dataset from UMD DRUM, which fails more often than not – DRUM seems to be very unreliable lately. I added an example that uses the dataset. We could exclude the example from testing, but I like the example in the DOCS for boosting to point to ...

.github/workflows/python-package-conda.yml
```diff
@@ -13,24 +13,30 @@ jobs:
...
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python 3.10
-      uses: actions/setup-python@v3
-      with:
-        python-version: '3.10'
+    # - name: Set up Python 3.10
+    #   uses: actions/setup-python@v3
+    #   with:
+    #     python-version: '3.10'
```
christianbrodbeck: Is there a potential use for these comments or can they be deleted?

.github/workflows/python-package-conda.yml
```diff
@@ -13,24 +13,30 @@ jobs:
...
       with:
-        environment-file: env-dev.yml
         environment-name: eeldev
+        create-args: >-
+          python=3.10
+          libcblas
+          numpy>=1.20
+          wxpython>=4.1.0
+          rpy2
+          r-car
```
christianbrodbeck: Is it not advisable to use `env-dev` here? I'm just wondering whether we can minimize the number of places where the dependencies have to be updated (e.g. if the required `numpy` version changes).

christianbrodbeck (commented):
Thanks!! Just two small questions.

.github/workflows/python-package-conda.yml
proloyd: Yes, I will remove them. This was a draft to see your thoughts.

.github/workflows/python-package-conda.yml
proloyd: Certainly. Or we could just have separate yml files for github actions. Your call.
But, now you can have a much smaller env-dev.file, since `pyproject.toml` specifies most of the dependencies, even for docs and all.

Another thing:
this is the minimal set of package that needs to be installed using `conda`, before we can successfully do `pip install .` .

proloyd:
> Yes, it's the download of the Alice dataset from UMD DRUM, which fails more often than not – DRUM seems to be very unreliable lately. I added an example that uses the dataset. We could exclude the example from testing, but I like the example in the DOCS for boosting to point to ...

I tried replacing the `urlretrieve` with a more reliable `pooch.retrieve`! But still it's the same issue on the docs runner. However on my macbook it works without any fuss!

.github/workflows/python-package-conda.yml
christianbrodbeck: Packages from `env` are installed from `conda-forge`, whereas packages in `pyproject.toml` are installed from `pip`, right? Back when I started with this there was an advantage to install as much as possible from `conda-forge` because that makes sure packages are all built with mutual compatibility. Is that less of an issue now?

christianbrodbeck:
I also tried to implement a download that can resume after an interruption and it did not help ....

.github/workflows/python-package-conda.yml
proloyd: >> Packages from env are installed from conda-forge, whereas packages in pyproject.toml are installed from pip.

Correct.

>> Is that less of an issue now?

It may be. And lots of big players (`PyTorch`) are dropping conda-forge support since pip has been more reliable. However, totally abandoning conda build is also not sustainable. For the timing, we can move these packages to `env-dev.yaml`. By the way, my intention on having a minimal set of packages (that needs to be installed from conda) is to keep ourselves more vigilant on something not working.

proloyd: For the time being, I will just go with `eel-dev.yml` to create the environment, since we are following similar strategy for readthedocs.

eelbrain/datasets/_alice.py
```diff
@@ -1,40 +1,42 @@
...
-import shutil
+import os
+import pooch
+
 from pathlib import Path
-from urllib.request import urlretrieve
 import zipfile
 
-from pooch import file_hash
-
```
christianbrodbeck, line 9: Small style request
```
Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.

You should put a blank line between each group of imports.
``` 
(https://peps.python.org/pep-0008/)

eelbrain/datasets/_alice.py
```diff
@@ -1,40 +1,42 @@
...
 
 
 def get_alice_path(
         path: PathArg = Path("~/Data/Alice"),
 ):
-    md5 = hashlib.md5()
-
     path = Path(path).expanduser().resolve()
-    if path.exists():
-        return path
```
christianbrodbeck, line 20: Deleting this will cause the function to download the dataset even if it's already downloaded.

eelbrain/datasets/_alice.py
```diff
@@ -1,40 +1,42 @@
...
+    }
+    fetcher = pooch.Pooch(
+        path=path,
+        base_url=baseurl,
+        registry=registry,
+        retry_if_failed=4,
+    )
+    for fname in registry.keys():
+        fetcher.fetch(fname, progressbar=True, processor=pooch.Unzip(extract_dir='.'))
+        os.remove(path / fname)
```
christianbrodbeck: In case you were looking for this in `pathlib`, it's `Path.unlink()` :)

eelbrain/datasets/_alice.py
```diff
@@ -1,40 +1,42 @@
...
-        # if file_hash != hash:
-        #     raise RuntimeError(f'Hash mismatch for {url}: {file_hash} != {hash}')
-        with zipfile.ZipFile(temp_file_path, 'r') as f:
-            f.extractall(path)
-        Path(temp_file_path).unlink()
+    # for url, known_hash in urls:
+        # temp_file_path = pooch.retrieve(url, known_hash, progressbar=True)
+        # with zipfile.ZipFile(temp_file_path, 'r') as f:
+        #     f.extractall(path)
+        # os.remove(temp_file_path)
```
christianbrodbeck: Delete comment?

eelbrain/datasets/_alice.py
proloyd: But, if the previous download failed, then it will just proceed with that. Let me find an workaround with this.

proloyd: Now I check if the sub-directories exist before fetching the zip files.

proloyd:
> I also tried to implement a download that can resume after an interruption and it did not help ....

This is a major pain point. DRUM is very finicky. Is there any workaround? Since you use only one subject's data, could you not use a separate github repo to store them and use it from there?

eelbrain/datasets/_alice.py
```diff
@@ -1,40 +1,32 @@
...
-        return path
-    path.mkdir(exist_ok=True, parents=True)
-    urls = [
-        ['https://drum.lib.umd.edu/bitstream/handle/1903/27591/stimuli.zip', '4336a47bef7d3e63239c40c0623dc186'],
-        # ['https://drum.lib.umd.edu/bitstream/handle/1903/27591/eeg.0.zip', 'd63d96a6e5080578dbf71320ddbec0a0'],
-        ['https://drum.lib.umd.edu/bitstream/handle/1903/27591/eeg.1.zip', 'bdc65f168db4c0f19bb0fed20eae129b'],  # S15-S34
-        # ['https://drum.lib.umd.edu/bitstream/handle/1903/27591/eeg.2.zip', '3fb33ca1c4640c863a71bddd45006815'],
-    ]
+    if not path.exists():
+        path.mkdir(exist_ok=True, parents=True)
```
christianbrodbeck:
```suggestion
    path.mkdir(exist_ok=True, parents=True)
```
`exist_ok=True` takes care of the conditional

christianbrodbeck:
Wow, 2 out of 3 succeeded :) 

Thanks @proloyd !!

PR #88, opened 2025-05-23
fix: replace connectivity with adjacency in all variables/function names to align with MNE-Python

zhangwenchili:


eelbrain/_experiment/mne_experiment.py
```diff
@@ -6727,12 +6727,12 @@ def _set_analysis_options(
...
-            # source connectivity
-            connectivity = self.get('connectivity')
-            if connectivity and not data.source:
-                raise NotImplementedError(f"connectivity={connectivity!r} is not implemented for data={data!r}")
-            elif connectivity:
-                items.append(connectivity)
+            # source adjacency
+            adjacency = self.get('adjacency')
+            if adjacency and not data.source:
+                raise NotImplementedError(f"adjacency={adjacency!r} is not implemented for data={data!r}")
```
```suggestion
                raise NotImplementedError(f"{adjacency=} is not implemented for {data=}")
```
I'm updating those whenever I come across them

eelbrain/mne_fixes/_version.py
```diff
@@ -11,7 +11,9 @@
 V1 = packaging.version.parse('1')
 
 # 0.22 renamed read_ch_connectivity → read_ch_adjacency
-if MNE_VERSION < V0_22:
-    mne.channels.find_ch_adjacency = mne.channels.find_ch_connectivity
-    mne.channels.read_ch_adjacency = mne.channels.read_ch_connectivity
-    mne.spatial_src_adjacency = mne.spatial_src_connectivity
+# UPDATE: we don't support mne < 1 anymore
```
christianbrodbeck: You can just delete these lines, no need to keep a comment, the requirement is specified in [`pyproject.toml`](https://github.com/Eelbrain/Eelbrain/blob/e38452d80fa643721b5d2cac516e9809f684d8c6/pyproject.toml#L31)

christianbrodbeck (changes requested):
Just a small change. Thanks!

eelbrain/mne_fixes/_version.py
zhangwenchili: Thanks for pointing this out! Deleted in 3b5f58f35457137cc3e3ed381c608e6f033cca5e.

eelbrain/_experiment/preprocessing.py
```diff
@@ -253,8 +253,8 @@ def _as_dict(self, args: Sequence[str] = ()):
                 out['montage'] = Sensor.from_montage(self.montage)
             else:
                 out['montage'] = self.montage
-        if self.connectivity is not None:
-            out['connectivity'] = self.connectivity
+        if self.adjacency is not None:
+            out['adjacency'] = self.adjacency
```
christianbrodbeck: This is another backwards compatibility issue. We should just keep the dictionary key. In the preprocessing module, the `._as_dict()` dictionary is used to compare the current settings with the pickled settings from previous runs (https://github.com/Eelbrain/Eelbrain/blob/2d62540fa13ac55fe38fc6179d1214ce033329e6/eelbrain/_experiment/mne_experiment.py#L933), so a key change will lead to spurious differences. We could later change the key if we get to a point where we can't be backwards compatible (like when changing data input structure)

christianbrodbeck (changes requested):

eelbrain/_experiment/preprocessing.py
zhangwenchili: Thanks for pointing this out! Changed in fe189d67c57e8a3f43322398e144952ad6c7aaa0. Actucally this is what I'm trying to ask in https://github.com/zhangwenchili/Eelbrain/pull/6#discussion_r2104696475, but I made a mistake by not quoting the dictionary key there :)

After changing this, I searched `adjacency'` in the repo again. It occurred in two files, one is a field in the `MneExperiment` class, and the other is here

image

Will they create backward compatibility problems as well?

christianbrodbeck: Good thinking! You're right, this `__getstate__` and `__setstate__` pair for `NDPermutationDistribution` has the same issue.

zhangwenchili: Fixed in 9845a4e9265d934644e81dbb37fd956ce57209c7.

zhangwenchili: I'm trying to be more careful with this. Which one do you think is better for `NDPermutationDistribution.__getstate__`?
```
state = {
            name: getattr(self, name) for name in (
                'name', 'meas', '_version', '_host', '_init_time',
                # settings ...
                'kind', 'threshold', 'tfce', 'tail', 'criteria', 'samples', 'tstart', 'tstop', 'parc',
                # data properties ...
                'dims', 'shape', '_nad_ax', '_vector_ax', '_criteria', '_connectivity',
                # results ...
                'dt_original', 'dt_perm', 'n_clusters', '_dist_dims', 'dist', '_original_param_map', '_original_cluster_map', '_cids',
            )}
```
or
```
state = {
            name: getattr(self, name) for name in (
                'name', 'meas', '_version', '_host', '_init_time',
                # settings ...
                'kind', 'threshold', 'tfce', 'tail', 'criteria', 'samples', 'tstart', 'tstop', 'parc',
                # data properties ...
                'dims', 'shape', '_nad_ax', '_vector_ax', '_criteria',
                # results ...
                'dt_original', 'dt_perm', 'n_clusters', '_dist_dims', 'dist', '_original_param_map', '_original_cluster_map', '_cids',
            )}
state['_connectivity'] = self._adjacency
```
The first one doesn't rename self._connectivity while the second do :)

christianbrodbeck: I could see an argument for either one. Maybe a slight preference for renaming the attribute, as it would make code easier to read in the future?

zhangwenchili: Thanks for the suggestion! I'll use the second one then.

eelbrain/_stats/testnd.py
```diff
@@ -3242,7 +3243,10 @@ def __setstate__(self, state):
             state['tfce'] = ['kind'] == 'tfce'
 
         for k, v in state.items():
```
christianbrodbeck, line 3246:
```suggestion
        self._adjacency = state.pop('_connectivity')
        for k, v in state.items():
```
Just a small STY suggestion, otherwise it all looks good!

eelbrain/_stats/testnd.py
```diff
@@ -3242,7 +3243,10 @@ def __setstate__(self, state):
...
-            setattr(self, k, v)
+            if k == '_connectivity':
+                self._adjacency = v
+            else:
+                setattr(self, k, v)
```
christianbrodbeck:
```suggestion
            setattr(self, k, v)
```

eelbrain/_stats/testnd.py
zhangwenchili: This looks much better, thanks!

christianbrodbeck:
Thanks @zhangwenchili !

PR #102, opened 2025-07-11
FIX:  xhemi() for partial source space

christianbrodbeck:
@jpereznavarro this replicates your error, now working on a fix


christianbrodbeck:
@jpereznavarro can you check whether this branch fixes your issue?

PR #104, opened 2025-07-26
FIX space in channel names

zhangwenchili:
Maybe we can instead just change the condition. Looks like only neuromag systems have this issue.

### DIFF
--- a/eelbrain/_io/fiff.py
+++ b/eelbrain/_io/fiff.py
@@ -784,7 +784,7 @@ def sensor_dim(
             # fix channel names
             if adjacency.startswith('neuromag'):
                 if (' ' not in adj_ch_names[0]) and (' ' in ch_names[0]):  # mne-python < ~1.2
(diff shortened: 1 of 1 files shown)


eelbrain/_io/fiff.py
```diff
@@ -783,19 +783,14 @@ def sensor_dim(
             c_matrix, adj_ch_names = mne.channels.read_ch_adjacency(adjacency)
             # fix channel names
             if adjacency.startswith('neuromag'):
-                if ' ' not in adj_ch_names[0]:  # mne-python < ~1.2
+                if (' ' not in adj_ch_names[0]) and (' ' in ch_names[0]):  # mne-python < ~1.2
```
christianbrodbeck: Remove the comment, because it looks like it's inconsistent in some datasets, not just mne-Python
```suggestion
                if (' ' not in adj_ch_names[0]) and (' ' in ch_names[0]):
```

christianbrodbeck:
Good catch! Thanks @zhangwenchili !

PR #112, opened 2025-09-23
FIX: CapWords for class name

zhangwenchili:
1. MneExperiment -> Pipeline
2. Fix `CachedRawPipe.cache()`


christianbrodbeck:
What about naming it `Pipeline` instead of `MNEExperiment`?

zhangwenchili:
> What about naming it `Pipeline` instead of `MNEExperiment`?

Updated in d2da4b8102d1591bf8fd27e6fd74c1cbe0a48d25.

christianbrodbeck:
Thanks @zhangwenchili !!

PR #114, opened 2025-10-09
Fix mri subject name, datatype and automatically create channels.tsv

zhangwenchili:
- [x] https://github.com/zhangwenchili/Eelbrain/issues/25
- [x] https://github.com/zhangwenchili/Eelbrain/issues/26
~~- [ ] https://github.com/zhangwenchili/Eelbrain/issues/27~~

### DIFF
--- a/doc/experiment.rst
+++ b/doc/experiment.rst
@@ -71,8 +71,8 @@ The pipeline expects input dataset in `BIDS (Brain Imaging Data Structure) <http
     root                              {root}
     subject folder                       /sub-{subject}
     session folder                          /ses-{session}
-    modality folder                            /{modality}
-    raw data file                                 /sub-{subject}_ses-{session}_task-{task}_run-{run}_{modality}.fif
+    datatype folder                            /{datatype}
+    raw data file                                 /sub-{subject}_ses-{session}_task-{task}_run-{run}_{datatype}.fif
     derivatives root                     /derivatives
     trans file                              /trans/sub-{subject}_ses-{session}_trans.fif
     FreeSurfer MRI                          /freesurfer
@@ -83,7 +83,7 @@ The pipeline expects input dataset in `BIDS (Brain Imaging Data Structure) <http
     In BIDS specification, ``{root}/derivatives`` is for files that do not fit into the BIDS structure, such as FreeSurfer MRIs and Eelbrain-generated files.
 
 
-``{subject}``, ``{session}``, ``{task}``, ``{run}`` and ``{modality}`` are `BIDS entities <https://bids-specification.readthedocs.io/en/stable/appendices/entities.html>`_. ``{session}`` and ``{run}`` are optional. ``{modality}`` is ``meg`` by default.
+``{subject}``, ``{session}``, ``{task}`` and ``{run}`` are `BIDS entities <https://bids-specification.readthedocs.io/en/stable/appendices/entities.html>`_. ``{session}`` and ``{run}`` are optional.
 
 
 ``MRI`` files (including ``trans-file``) are optional and only needed for source localization. The ``{root}/derivatives/freesurfer`` directory is `FreeSurfer <https://surfer.nmr.mgh.harvard.edu>`_ subject directory. They either contain the files created by FreeSurfer's `recon-all <https://surfer.nmr.mgh.harvard.edu/fswiki/recon-all>`_ command, or are created by the MNE-Python coregistration utility for scaled template brains. (Note that the pipeline doesn't use the NIfTI format that BIDS specifies.) A corresponding ``trans-file`` is created with the MNE-Python coregistration utility in either case (see more information on using `structural MRIs <https://github.com/christianbrodbeck/Eelbrain/wiki/Coregistration%3A-Structural-MRI>`_ or the `fsaverage template brain <https://github.com/christianbrodbeck/Eelbrain/wiki/Coregistration%3A-Template-Brain>`_).
@@ -130,7 +130,7 @@ IN the example above, the following would be saved in ``~/Code/MyProject/my_expe
 
     class MyExperiment(Pipeline):
 
-        modality = 'eeg'
+        datatype = 'eeg'
 
     e = MyExperiment("~/Data/Experiment")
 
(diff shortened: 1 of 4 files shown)


doc/experiment.rst
```diff
@@ -71,8 +71,8 @@ The pipeline expects input dataset in `BIDS (Brain Imaging Data Structure) <http
     root                              {root}
     subject folder                       /sub-{subject}
     session folder                          /ses-{session}
-    modality folder                            /{modality}
-    raw data file                                 /sub-{subject}_ses-{session}_task-{task}_run-{run}_{modality}.fif
+    datatype folder                            /{datatype}
+    raw data file                                 /sub-{subject}_ses-{session}_task-{task}_run-{run}_{datatype}.fif
     derivatives root                     /derivatives
     trans file                              /trans/sub-{subject}_ses-{session}_trans.fif
     FreeSurfer MRI                          /freesurfer
```
christianbrodbeck:
```suggestion
    FreeSurfer SUBJECTS_DIR                 /freesurfer
    FreeSurfer subject                         /sub-{subject}
```

christianbrodbeck: There is one inconsistency, which is that `fsaverage` will not be called `sub-fsaverage`. Will this cause a problem?

doc/experiment.rst
christianbrodbeck:
```suggestion
- Defining a :class:`Pipeline` subclass with the parameters required to find those files
```

christianbrodbeck:
```suggestion
In the example above, the following would be saved in ``~/Code/MyProject/my_experiment.py``::
```

doc/experiment.rst
```diff
@@ -130,7 +130,7 @@ IN the example above, the following would be saved in ``~/Code/MyProject/my_expe
 
     class MyExperiment(Pipeline):
 
-        modality = 'eeg'
+        datatype = 'eeg'
```
christianbrodbeck: Shouldn't `datatype` really be a state parameter? If there are multiple datatypes, the user will probably want to analyze both, so they should be able to switch back and forth. If there is just one, then we can infer what the user wants, so they should not need worry about it.

For now, I think you could just leave the code as is (giving a `NotImplementError` when there are more than one data types, and with the advanced option of setting it manually) but not mention it during the introduction/easy setup?

christianbrodbeck (changes requested):

christianbrodbeck:
@zhangwenchili If the subjects name issue is not straight forward we could merge this before addressing the subject name so Aditya can keep working

doc/experiment.rst
zhangwenchili: Fixed in 2a2edd9f8d3fd908a7f084c5f639595befff76d2. Yes, it needs to be handled separately. Let's move the discussion to https://github.com/zhangwenchili/Eelbrain/issues/27.

doc/experiment.rst
zhangwenchili: Fixed in https://github.com/Eelbrain/Eelbrain/commit/2a2edd9f8d3fd908a7f084c5f639595befff76d2.

doc/experiment.rst
zhangwenchili: Fixed in https://github.com/Eelbrain/Eelbrain/commit/2a2edd9f8d3fd908a7f084c5f639595befff76d2.

If we allow switching datatype after init, the value list (second parameter of `_register_field`) of `task` and `run` should depend on `datatype`. Currently there is no well implemented way to do this. Will it be dangerous if I do `_register_field` in `post_set_handler`?

zhangwenchili:
> @zhangwenchili If the subjects name issue is not straight forward we could merge this before addressing the subject name so Aditya can keep working

Sure!

doc/experiment.rst
christianbrodbeck: Ah, right, that makes it slightly more complicated. It should be straight forward to add using these handlers.

doc/experiment.rst
```diff
@@ -63,27 +63,27 @@ Setting up the file structure
...
     subject folder                       /sub-{subject}
     session folder                          /ses-{session}
-    modality folder                            /{modality}
-    raw data file                                 /sub-{subject}_ses-{session}_task-{task}_run-{run}_{modality}.fif
+    datatype folder                            /{datatype}
+    raw data file                                 /sub-{subject}_ses-{session}_task-{task}_run-{run}_{datatype}.fif
     derivatives root                     /derivatives
     trans file                              /trans/sub-{subject}_ses-{session}_trans.fif
-    FreeSurfer MRI                          /freesurfer
+    FreeSurfer SUBJECTS_DIR                          /freesurfer
```
christianbrodbeck:
```suggestion
    FreeSurfer SUBJECTS_DIR                 /freesurfer
```

doc/experiment.rst
```diff
@@ -461,10 +461,10 @@ Exclude certain entities from the experiment, e.g.::
         'session': ['emptyroom'],
     }
 
-.. py:attribute:: Pipeline.modality
-   :type: str
+.. .. py:attribute:: Pipeline.datatype
+..    :type: str
```
christianbrodbeck, line 465: Are these double dots intentional?

eelbrain/_experiment/preprocessing.py
```diff
@@ -88,14 +88,17 @@ def get_path(
             self,
             path: BIDSPath,
             file_type: Literal['raw', 'bads', 'cache', 'ica'] = 'raw',
-            on_error: Literal['raise', 'warn', 'ignore'] = 'raise',
-    ) -> str | None:
-        if not path.find_matching_sidecar(on_error=on_error):
-            return None
+    ) -> str:
+        # Caller of this method should check raw exists or not. Channels file will be created if it doesn't exist.
```
```suggestion
        # Caller of this method should check whether raw exists or not. 
        # Channels file will be created if it doesn't exist.
```

christianbrodbeck (commented):
Just some minor issues/questions

doc/experiment.rst
zhangwenchili: `.. ` is commenting out these lines. I'm leaving the old approach as comments until the new one is implemented.

christianbrodbeck:
Thanks @zhangwenchili !

PR #117, opened 2025-10-20
Fix subject name and bad channels

zhangwenchili:


eelbrain/_experiment/mne_experiment.py
```diff
@@ -5552,7 +5552,7 @@ def make_report_coreg(self, file_name=None, **state):
                                            brain_surfaces='white', show=False)
 
             # add to report
-            if subject == mrisubject:
+            if 'sub' + subject == mrisubject:
```
```suggestion
            if 'sub-' + subject == mrisubject:
```
Should this have a dash?

christianbrodbeck (commented):
Just saw one potential issue, otherwise looks good, were you able to test it in some form?

christianbrodbeck:
@zhangwenchili is this ready to be combined with the re-reference fix in the release?

eelbrain/_experiment/mne_experiment.py
zhangwenchili: Fixed in 3faa0b7fe10722504c2185a56ed1f795aa570ba4

zhangwenchili:
@christianbrodbeck Now it's ready.

zhangwenchili:
> were you able to test it in some form

For now I did some manual tests on the two datasets.
For the ImageNet dataset:
```
root = '/mnt/d/Data/ds005810'
e = ImageNet('/mnt/d/Data/ds005810')
e.set(rej='', raw='1-40')
for _ in e.iter():
    print(e.load_evoked_stc())
print(e.load_test('connection', 0.3, 0.5, 0.05, data='meg', baseline=False, make=True))
```
For the Aditya's dataset:
```
root = '/mnt/d/Data/story'
e = Story(root)
e.set(raw='ica', rej='')
e.make_ica()
e.make_ica_selection()
print(e.load_events())
```
As we discussed on Tuesday, I'm adding some unit tests for preprocessing classes, which will make future testing more reliable.

eelbrain/_experiment/preprocessing.py
```diff
@@ -314,9 +328,10 @@ def cache(self, path: BIDSPath) -> None:
         return None
 
     def _load(self, path: BIDSPath, preload):
```
christianbrodbeck: Can you complete type hints?

eelbrain/_experiment/preprocessing.py
```diff
@@ -511,14 +526,14 @@ def load(
...
+            preload: bool = False,
             raw: mne.io.BaseRaw = None,
     ) -> mne.io.BaseRaw:
         if raw is not None:
             pass
         elif self._cache:
             raw = self.cache(path, load=True, preload=preload)
-        elif preload == -1:
+        elif preload == False:
             raw = self._make_info(path)
```
christianbrodbeck: This is not the same, `_make_info` returns the `raw` of the parent pipe:

```python
    def _make_info(self, path: BIDSPath) -> mne.io.BaseRaw:
        return self.source.load(path, preload=False)
```

When you just set `preload=False` you may still want the `raw` of the current pipe, so that you have the option to later call `raw.load_data()`. 

The idea behind allowing `preload=-1` was that there are cases where you know you will never want to load the data for that `raw`, so in that case you can be lazy (not generate the raw data until it's actually needed).

If there is no more need for that lazy option then maybe support for it can be dropped entirely?

christianbrodbeck (changes requested):

eelbrain/_experiment/preprocessing.py
zhangwenchili: > you will never want to load the data for that raw

So this can be achieved by setting `raw='raw'` right? If that's the case I think it's better to drop it to avoid confusion.

christianbrodbeck: Not quite, in subclasses `_make_info` would update info without actually computing raw. For example:

```python
class RawFilter(CachedRawPipe):

    ...

    def _make_info(self, path: BIDSPath) -> mne.io.BaseRaw:
        raw = super()._make_info(path)
        l_freq, h_freq = self.args
        if l_freq and l_freq > (raw.info['highpass'] or 0):
            with raw.info._unlock():
                raw.info['highpass'] = float(l_freq)
        if h_freq and h_freq < (raw.info['lowpass'] or raw.info['sfreq']):
            with raw.info._unlock():
                raw.info['lowpass'] = float(h_freq)
        return raw
```

A cleaner and less confusing implementation would be a `.get_info()` method that constructs `info` only.

zhangwenchili: So `._make()` changes raw data and info, while `._make_info()` changes info only?

> A cleaner and less confusing implementation would be a .get_info() method that constructs info only.

Yes I think this is a better way. What about `.load_info()` to align with `.load_raw()`?

Also need to add `.load_info()` to `Pipeline` because currently there's no way to invoke it.

zhangwenchili: Updated in 8f80a226fed44bb98e025aa6128274a36b6558c4.

eelbrain/_experiment/preprocessing.py
zhangwenchili: Fixed in 8f80a226fed44bb98e025aa6128274a36b6558c4.

eelbrain/_experiment/preprocessing.py
christianbrodbeck: The pipes involving ICA would need a `load_info` override too, because they potentially combine bad channels from multiple raw files.

```python
class RawICA(CachedRawPipe):

    ...

    def _apply(
            self,
            raw: mne.io.BaseRaw,
            path: BIDSPath,
            raw_name: str,
    ) -> mne.io.BaseRaw:
        self.log.debug("Raw %s: applying ICA for %s...", raw_name, path.fpath)
        raw.info['bads'] = self.load_bad_channels(path, existing=raw.ch_names)
        ica = self.load_ica(path)
        missing = self._check_ica_channels(ica, raw, raise_on_mismatch=True, raw_name=raw_name, subject=path.entities['subject'], return_missing=True)
        if missing:
            raw.drop_channels(missing)
        ica.apply(raw)
        return raw
```

In order to avoid duplicating code (`raw.info['bads'] = self.load_bad_channels(path, existing=raw.ch_names)` in `.load_info()` and in `._apply()`) maybe the `RawPipe` should have a `._update_info` method that can be called from both?

zhangwenchili: Updated in 03d59b4883689f9d5887a1ce8fe9ef9d5472c728. Well I think a simple assignment does not worth abstraction?

eelbrain/_experiment/mne_experiment.py
```diff
@@ -3355,6 +3354,20 @@ def load_raw(
 
         return raw
 
+    def load_info(self, **kwargs) -> mne.Info:
```
christianbrodbeck: I would hesitate making this public API. I've used it internally with `info` fields I cared about, bit I am not sure if `info` is updated completely. If it's useful we can add it as private method.

eelbrain/_experiment/preprocessing.py
```diff
@@ -118,34 +131,38 @@ def load(
...
-            path,
-            extra_params={'preload': preload},
+        raw_path = self.get_path(path)
+        return mne.io.read_raw_fif(
+            raw_path,
+            preload=preload,
             verbose='critical',
         )
 
+    def load_info(self, path: BIDSPath) -> mne.Info:
```
christianbrodbeck, line 163: Can we add a one sentence docstring?

eelbrain/_experiment/preprocessing.py
```diff
@@ -855,7 +883,7 @@ def make_ica(
         ica_path = self.get_path(path, 'ica')
         bad_channels = self.load_bad_channels(path)
         if exists(ica_path):
-            raw = self.source.load(path.copy().update(task=self.task[0]), bad_channels, preload=-1)
+            raw = self.source.load(path.copy().update(task=self.task[0]), bad_channels)
```
christianbrodbeck: Here you'd now want to use `load_info()`. One common pattern is
```
for subject in pipeline:
    subject.make_ica()
```
In this case you just want to skip through subjects for which ICA already exists. `self.source.load()` can take much longer than `self.load_info()`, e.g. if the previous raw-pipe is not cached. That was the reason for having `preload=-1` here.

eelbrain/_experiment/preprocessing.py
```diff
@@ -1201,11 +1229,11 @@ def _make(
             raw = raw.drop_channels(self.drop)
         return raw
 
-    def _make_info(self, path: BIDSPath) -> mne.io.BaseRaw:
+    def load_info(self, path: BIDSPath) -> mne.Info:
         if self.add or self.drop:
-            return self._make(path, False)
+            return self._make(path, False).info
```
```suggestion
            return self.load(path).info
```
Use cache if available?

eelbrain/_experiment/preprocessing.py
christianbrodbeck, line 1025: `RawApplyICA.load_info()` should also update bad channels

christianbrodbeck (changes requested):

eelbrain/_experiment/preprocessing.py
zhangwenchili: Fixed in e5ce78a87b269a3195a95d19212c00bc663fe9d5

eelbrain/_experiment/preprocessing.py
zhangwenchili: Fixed in e5ce78a87b269a3195a95d19212c00bc663fe9d5

eelbrain/_experiment/preprocessing.py
zhangwenchili: Fixed in e5ce78a87b269a3195a95d19212c00bc663fe9d5

eelbrain/_experiment/mne_experiment.py
zhangwenchili: Changed in e5ce78a87b269a3195a95d19212c00bc663fe9d5

eelbrain/_experiment/preprocessing.py
zhangwenchili: `Raw` instead of `Info` is needed here, so there's two options:
1. Let `load_info()` return raw (`load_info()` -> `load_with_processed_info()`).
2. Ignore this case, since `make_ica()` and caching should normally be one-time operation.

christianbrodbeck: It should be fairly simple to rewrite `._check_ica_channels()` to take only `info` and not the full `Raw`.

(2. not necessarily – For example, you may often start preliminary analysis with a partial dataset and then add more subjects as you collect more data, then you would repeatedly run the `for` loop above to create ICA for the subjects you just added)

zhangwenchili: You're right, thanks! Fixed in 6b9dc4c93c7425419cb0b37673a95f6d2d387de7.

eelbrain/_experiment/preprocessing.py
```diff
@@ -853,12 +882,11 @@ def make_ica(
             run: list[str],
     ) -> str:
         ica_path = self.get_path(path, 'ica')
-        bad_channels = self.load_bad_channels(path)
         if exists(ica_path):
-            raw = self.source.load(path.copy().update(task=self.task[0]), bad_channels, preload=-1)
+            info = self.load_info(path.copy().update(task=self.task[0]))
```
christianbrodbeck: Shouldn't this be `self.source.load_info()`?

christianbrodbeck:
Thanks @zhangwenchili

PR #124, opened 2025-11-13
feature:  EEG input to BIDS pipeline

christianbrodbeck:
This enables EEG input in the BIDS allowed formats.

I also renamed the `T` event column to `time`.

### DIFF
--- a/eelbrain/_experiment/preprocessing.py
+++ b/eelbrain/_experiment/preprocessing.py
@@ -1263,7 +1263,7 @@ def assemble_pipeline(
                 if isinstance(pipe, RawICA):
                     missing = set(pipe.task).difference(tasks)
                     if missing:
-                        raise DefinitionError(f"RawICA {key!r} lists one or more non-exising tasks: {', '.join(missing)}")
+                        raise DefinitionError(f"RawICA {key!r} lists one or more non-exising tasks: {', '.join(missing)}. Available tasks: {', '.join(tasks)}.")
                 linked_raw[key] = pipe
         if len(raw) == n:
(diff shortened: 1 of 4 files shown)


eelbrain/_experiment/preprocessing.py
```diff
@@ -327,7 +327,20 @@ def _load(
...
+            case '.set':
+                reader = mne.io.read_raw_eeglab
+            case '.bdf':
+                reader = mne.io.read_raw_bdf
+            case _:
+                raise RuntimeError(f"Unrecognized file format: {path.suffix}")
+        raw = reader(
             raw_path,
             preload=preload,
             verbose='critical',
```
zhangwenchili, line 346: We have two ._load(). Maybe move this to Rawpipe._load() and call Rawpipe._load() here?

zhangwenchili (commented):
Others look good. Did you try this on a EEG dataset?

eelbrain/_experiment/preprocessing.py
christianbrodbeck: The reason for having two `._load()` implementations is that `RawPipe._load` is for cached data which is always in FIFF format. `RawSource` overloads it to handle more complex input data.

christianbrodbeck:
Thanks @zhangwenchili ! 

Yes, I tested it with an EEG dataset (unfortunately not a public one so it's not suitable for a test at this point)

eelbrain/_experiment/preprocessing.py
zhangwenchili: Thanks for the reminder!

PR #128, opened 2025-11-19
FIX glob pattern

### LINKED ISSUE #31: Update parameter name for `PrimaryEpoch`

### LINKED ISSUE #127: BIDS:  glob pattern wrong
I get:
```python
>>> e._glob_pattern('ica-file', inclusive=True)
'/Users/christian/Data/SLI-in-school/derivatives/ica/_raw-*_ica.fif'
```

I would have expected:

```python
'/Users/christian/Data/SLI-in-school/derivatives/ica/*_raw-*_ica.fif'
```

Because the path template is:

```
'ica-file': join('{deriv-dir}', 'ica', '{epoch_basename}_raw-{raw}_ica.fif'), 
``` 

@zhangwenchili can you have a look?

zhangwenchili:
- Closes #127 
- Closes https://github.com/zhangwenchili/Eelbrain/issues/31

### DIFF
--- a/eelbrain/_experiment/epochs.py
+++ b/eelbrain/_experiment/epochs.py
@@ -88,7 +88,7 @@ class Epoch(EpochBase):
 
     # to be set by subclass
     rej_file_epochs = None
-    sessions = None
+    tasks = None
 
     def __init__(self, tmin=-0.1, tmax=0.6, samplingrate=None, decim=None, baseline=None,
                  vars=None, trigger_shift=0., post_baseline_trigger_shift=None,
@@ -157,8 +157,8 @@ class PrimaryEpoch(Epoch):
 
     Parameters
     ----------
-    session : str
-        Session (raw file) from which to load data.
+    task : str
+        Task (raw file) from which to load data.
     sel : str
         Expression which evaluates in the events Dataset to the index of the
         events included in this Epoch specification.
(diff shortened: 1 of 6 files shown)


christianbrodbeck (changes requested):
```py
>>> e._glob_pattern('ica-file', subject='2001', inclusive=True)
'/Users/christian/Data/SLI-in-school/derivatives/ica/*_raw-*_ica.fif'
```

I think I would expect:

```py
'/Users/christian/Data/SLI-in-school/derivatives/ica/sub-2001*_raw-*_ica.fif'
```

Relatedly, I think it would be good to add some tests for the `_glob_pattern` method?

zhangwenchili:
> I think I would expect:
> 
> ```python
> '/Users/christian/Data/SLI-in-school/derivatives/ica/sub-2001*_raw-*_ica.fif'
> ```

Current template don't return this because `{epoch_basename}` is a single field.
`'ica-file': join('{deriv-dir}', 'ica', '{epoch_basename}_raw-{raw}_ica.fif')`
Do you need this behaviour in some cases?

The reason for `{epoch_basename}` is that our path templating method doesn't handle optional fields like `session` and `run`. A cleaner way would be manually implement optional fields handling and abandon mne-bids. I think you mentioned the idea of only using mne-bids in raw data yesterday, but cached files also need these optional fields.

christianbrodbeck:
In a short run, what I would expect is that `epoch_basename` get set to `"sub-2001*"`. Would that cause any problems downstream? From a superficial view, it seems like it could be easily implemented in https://github.com/Eelbrain/Eelbrain/pull/128/files#diff-daf7dda682641b2596cba631902d9a9cc09b4bb1d1ae8059c4536d3010a3125cL6486

In the long run, yes it would be more desirable to somehow unify the two templating mechanisms. I think you're the one with the most experience with the two, what is your recommendation? :)

zhangwenchili:
> In a short run, what I would expect is that `epoch_basename` get set to `"sub-2001*"`. Would that cause any problems downstream? From a superficial view, it seems like it could be easily implemented in https://github.com/Eelbrain/Eelbrain/pull/128/files#diff-daf7dda682641b2596cba631902d9a9cc09b4bb1d1ae8059c4536d3010a3125cL6486

Not easy if using mne-bids, I made a solution of creating templates dynamically based on entities found. https://github.com/Eelbrain/Eelbrain/pull/128/commits/8604b1315f10211791477562a2f8753b5431a07e

Also I think you actually want:
`'/Users/christian/Data/SLI-in-school/derivatives/ica/sub-2001_ses-*_raw-*_ica.fif'`
if you have multiple sessions.


> In the long run, yes it would be more desirable to somehow unify the two templating mechanisms. I think you're the one with the most experience with the two, what is your recommendation? :)

I think manually implement optional fields handling based our template string approach is better, because mne-bids is for end users and doesn't provide much flexibility to extend the BIDS format.

christianbrodbeck:
Great, this works. Could we add a test? For example, testing:
```
e._glob_pattern('ica-file', inclusive=True)
e._glob_pattern('ica-file', subject='2001', inclusive=True)
```
etc.

zhangwenchili:
> Great, this works. Could we add a test? For example, testing:
> 
> ```
> e._glob_pattern('ica-file', inclusive=True)
> e._glob_pattern('ica-file', subject='2001', inclusive=True)
> ```
> 
> etc.

https://github.com/Eelbrain/Eelbrain/pull/128/commits/66285f58c293fa48d6eb07e031cc866fef687341

christianbrodbeck:
Thanks @zhangwenchili !!

PR #133, opened 2025-11-22
MAINT: Add release and dependabot config

larsoner:
Pulled from #132. Working on that PR will be a pain having to click "Approve workflows" every time so let's get in a trivial (and separable anyway!) part, then I'll keep working on #132.


eelbrain/plot/tests/test_brain.py
```diff
@@ -8,6 +9,12 @@
 from eelbrain.testing import hide_plots, requires_mne_testing_data
 
 
+pytestmark = pytest.mark.skipif(
+    os.getenv("GITHUB_ACTIONS") == "true" and sys.platform == "darwin",
+    reason="TODO: Segfaults on macOS GitHub Actions",
+)
```
larsoner, line 16: I looked at the diff in the `conda` envs and there wasn't anything suspicious, so I suspect this is actually from the transition from macos-13 (intel) to macos-14 (arm64) hardware default for `macos-latest`. But I think that can be investagated further separately, in the meantime this PR gets things green again @christianbrodbeck

.github/workflows/python-package-conda.yml
```diff
@@ -25,6 +30,9 @@ jobs:
         init-shell: bash
         cache-environment: true
         post-cleanup: 'all'
+        # Speed up env creation
+        create-args: |
+          mayavi==4.8.3 vtk==9.4.2 -v
```
larsoner: I noticed in https://github.com/Eelbrain/Eelbrain/pull/132 these were used, but sometimes it takes 10+ minutes for `micromamba` to resolve the env. I assume it's because of these deps. I'm also going to try adding `python=3.13` since those envs install `3.12.12`, which suggests some dependency is very out of date...

larsoner: Oh... it's `rpy2 < 3.6.1`

.github/workflows/python-package-conda.yml
```diff
@@ -36,8 +44,12 @@ jobs:
       run: |
         Xvfb :99 -screen 0 1024x768x16 &
         pytest eelbrain --no-gui
-    - if: matrix.os == 'macos-latest'
-      name: Test with GUIs and code-cov
```
larsoner, line 54: I don't see a reason in principle not to test on Linux as well (?), so I pushed a commit to setup a headless display with OpenGL support there (same action we use in MNE-Python etc.) and we'll see if it works... if it doesn't then I'll revert and let it test on the two macOS variants

larsoner: So it turns out those tests appeared to hang as soon as it was supposed to get to the `experiments` tests on both `macos-15-intel` and `ubuntu-latest`, not sure why. So I've gone back to just running GUI tests on `macos-latest`, but pushing `codecov` in both cases (it should be smart enough to merge coverages from multiple sources)

larsoner:
Oh wait... I realized why those tests are taking forever. They're accidentally downloading the `sample` dataset, which can take a variable amount of time (and quite a bit of it). I'll fix the skip here

larsoner:
Okay this one is green!

.github/workflows/python-package-conda.yml
```diff
@@ -25,24 +34,34 @@ jobs:
         init-shell: bash
         cache-environment: true
         post-cleanup: 'all'
+        # Speed up env creation (TODO: 3.12 because of rpy2 pin at least)
+        create-args: |
+          mayavi=4.8.3 vtk=9.4.2 python=3.12 -v
```
christianbrodbeck: Do we need to pin these? What does the TODO exactly refer to?

eelbrain/testing/_testing.py
```diff
@@ -197,14 +197,14 @@ def requires_framework_build(function):
 
 
 def requires_mne_sample_data(function):
-    if mne.datasets.sample.data_path(download=False):
```
larsoner, line 200: FYI this always returned `True`-like because if the dataset doesn't exist it returned `Path(".")`

larsoner:
This has conflicts... worth me fixing them for potential merge or should we close?

christianbrodbeck:
Yes, sorry for the delay! I just trimmed down the test env a bit, which should address your `TODO` regarding `rpy2` dependency.

larsoner:
Okay fixed the conflicts @christianbrodbeck , should be good to go

christianbrodbeck:
@codex review

eelbrain/datasets/_simple.py
```diff
@@ -141,7 +143,7 @@ def get_mne_evoked(ndvar=False):
         return evoked
 
 
-def get_mne_stc(ndvar=False, src='ico-5', subject='sample'):
+def get_mne_stc(ndvar=False, src='ico-5', subject='sample', *, download=False):
```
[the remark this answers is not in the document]
larsoner: Silently downloading the (huge) dataset is a bad thing. Better to let people get a failure that's easily fixed and let them choose to download it. 

Same with CIs, would rather code them explicitly to download somewhere than implicitly, silently do it.

.github/workflows/python-package-conda.yml
```diff
@@ -30,31 +38,44 @@ jobs:
         init-shell: bash
         cache-environment: true
         post-cleanup: 'all'
+        # Speed up env creation by pinning a couple of packages
+        # Limit to NumPy 2.2.6 to avoid in1d -> isin removal
+        create-args: |
+          mayavi=4.8.3 vtk=9.4.2 python=3.13 noqt5 pyside6=6.9.2 pysurfer=0.11.2 numpy=2.2.6 -v
```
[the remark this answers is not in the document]
larsoner: This one I should fix!

env-dev.yml
```diff
@@ -52,6 +52,7 @@ dependencies:
 - pingouin >= 0.6.0  # breaking API used in tests
 #- pytest-sugar  # errors on Linux
 # testing of R integration
+# TODO: This forces Python<3.13 because 3.6+ are first ones to support 3.13 on CF
```
christianbrodbeck: Can you remove `rpy2` (and `r` in general) from `env-dev`? I think it's niche enough now that if someone ever wants to use it they could install it separately.

.github/workflows/python-package-conda.yml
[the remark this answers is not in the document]
larsoner: ... fun fact about this, I don't think the *current* CI config actually installs 3.10. `matrix.python` isn't used anywhere. I'll fix it...

.github/workflows/python-package-conda.yml
```diff
@@ -30,31 +38,31 @@ jobs:
...
+        wm: false
     - name: Build eelbrain
       run: pip install --no-deps -e .
-      shell: bash -el {0}
     - name: Cache MNE datasets
       uses: actions/cache@v4
       with:
         path: ~/mne_data
-        key: mne-data-${{ matrix.os }}-${{ hashFiles('env-test.yml') }}
+        key: mne-data
```
larsoner, line 57: Unless you modify the test data somehow during tests (which would be bad!) and this depends on OS version (which would probably also be bad?) the key here should really only need to be the MNE version, not anything OS dependendent. And since you don't pin to a specific MNE version in the `env-test.yml`, the current setup will just lead to unnecessary cache misses and usage.

In practice the files you use don't really change from MNE version to MNE version, either, so you can just manually bump to `-1` etc. if needed at some point in the future.

env-test.yml
```diff
@@ -32,18 +32,15 @@ dependencies:
...
-- wxpython >= 4.0.3, != 4.1.0  # double install of libraries
+- pyside6
 # building
 - setuptools >= 17
 - cython >= 3
 # testing
-- pre-commit
 - pytest
 - pytest-cov
-- pytest-faulthandler
```
larsoner, line 43: Faulthandler support provided by `pytest` itself nowadays

larsoner (commented):
Okay @christianbrodbeck addressed AI feedback and all green 👍 

Would be good to merge this sooner rather than later probably so we don't end up duplicating effort again (as happened here setting up caching of datasets)

.github/workflows/python-package-conda.yml
```diff
@@ -30,33 +38,35 @@ jobs:
...
-        Xvfb :99 -screen 0 1024x768x16 &
-        pytest eelbrain --no-gui --no-mayavi
-    - if: matrix.os == 'macos-latest'
-      name: Test with GUIs and code-cov
-      shell: bash -el {0}
+        set -x
+        mne sys_info -d
+        time python -c "import mne; print(f'{mne.datasets.sample.data_path(verbose=False)=}')"
+        time python -c "import mne; print(f'{mne.datasets.testing.data_path(verbose=False)=}')"
+        time python -c "import mne; print(f'{mne.datasets.fetch_fsaverage(verbose=False)=}')"
```
christianbrodbeck, line 64: Are these `time` commands still needed?

eelbrain/_stats/tests/test_test.py
```diff
@@ -10,6 +9,8 @@
 from eelbrain.fmtxt import asfmtext
 from eelbrain.testing import assert_fmtxt_str_equals
 
+pingouin = pytest.importorskip('pingouin', minversion="0.6.0")
```
christianbrodbeck, line 12: Is this necessary if our `env-test` has `pingouin >= 0.6.0`?

christianbrodbeck (commented):
Just a few clarification points, thanks!

eelbrain/datasets/_simple.py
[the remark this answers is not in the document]
christianbrodbeck: Would it make sense to add a comment to `doc/development.rst` under [Testing](https://eelbrain.readthedocs.io/en/latest/development.html) that running the full test suite requires downloading the dataset?

.github/workflows/python-package-conda.yml
larsoner: Not needed, no. But I think they are nice to have for a little sense of how long each part takes when it does run, and there's no real downside as far as I see. This only ends up being < 20 lines so not overly verbose. (I tried `verbose=True` and that was thousands of lines, this seemed like a nicer compromise.)

eelbrain/_stats/tests/test_test.py
larsoner: If some person comes along and wants to tweak one tiny thing in `eelbrain` and runs `pytest eelbrain` (a reasonable thing to do) it's nice to have it just work by skipping tests that require `pingouin` and not error out on them. So to me it's a low cost way to lower the barrier to contribution. But no not totally necessary!

eelbrain/datasets/_simple.py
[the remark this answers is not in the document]
larsoner: Added some docs!

larsoner:
I keep seeing errors in the doc build like
```
../examples/temporal-response-functions/alice-trf.py unexpectedly failed to execute correctly:
    Traceback (most recent call last):
      File "/home/docs/checkouts/readthedocs.org/user_builds/eelbrain/checkouts/133/examples/temporal-response-functions/alice-trf.py", line 29, in <module>
        DATA_ROOT = eelbrain.datasets._alice.get_alice_path()
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/docs/checkouts/readthedocs.org/user_builds/eelbrain/conda/133/lib/python3.12/site-packages/eelbrain/datasets/_alice.py", line 29, in get_alice_path
        fetcher.fetch(fname, processor=pooch.Unzip(extract_dir='.'))
      File "/home/docs/checkouts/readthedocs.org/user_builds/eelbrain/conda/133/lib/python3.12/site-packages/pooch/core.py", line 598, in fetch
        stream_download(
      File "/home/docs/checkouts/readthedocs.org/user_builds/eelbrain/conda/133/lib/python3.12/site-packages/pooch/core.py", line 823, in stream_download
        downloader(url, tmp, pooch)
      File "/home/docs/checkouts/readthedocs.org/user_builds/eelbrain/conda/133/lib/python3.12/site-packages/pooch/downloaders.py", line 231, in __call__
        response.raise_for_status()
      File "/home/docs/checkouts/readthedocs.org/user_builds/eelbrain/conda/133/lib/python3.12/site-packages/requests/models.py", line 1168, in raise_for_status
        raise HTTPError(http_error_msg, response=self)
    requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://drum.lib.umd.edu/bitstream/handle/1903/27591/stimuli.zip
```
FWIW I've had (fairly) good luck with uploading data to osf.io and downloading it from there

larsoner:
Added a [section to the docs](https://eelbrain--133.org.readthedocs.build/en/133/development.html#testing-and-validation) and addressed comments!

christianbrodbeck:
@larsoner it looks like build-ubuntu-latest fails to resolve the test env with the new pins

larsoner:
Looking back, looks like it was actuallly removing the `pyside6` pin that broke things, probably became too complicated for the solver to deal with the variants of packages like `mayavi` that have qt5 and qt6 variants...

larsoner:
Okay actually green now!

christianbrodbeck (approved):
Thanks @larsoner !

PR #139, opened 2025-12-15
Add entity `split`, and refactor preprocessing

### LINKED ISSUE #137: Loading BIDS data with split raw files
Loading BIDS data with split raw files (Appleseed dataset)

- Add entity `split`
- resolve path starting with `~`

zhangwenchili:
- Closes #137. FIFF files in Appleseed dataset are too large, I only tested with `load_events(raw='raw')`
- Added unit tests for preprocessing.
- Refactored preprocessing to make code cleaner.

### DIFF
--- a/eelbrain/_experiment/mne_experiment.py
+++ b/eelbrain/_experiment/mne_experiment.py
@@ -2462,8 +2462,7 @@ class Pipeline(FileTree):
         elif data_raw:
             ds.info['raw'] = self.load_raw(add_bads, preload=self.preload)
 
-        for k, v in entities.items():
-            ds.info[k] = v
+        ds.info.update(entities)
 
         if self.trigger_shift:
             if isinstance(self.trigger_shift, dict):
--- a/eelbrain/_experiment/mne_experiment.py
+++ b/eelbrain/_experiment/mne_experiment.py
@@ -2187,7 +2187,6 @@ class Pipeline(FileTree):
         if ndvar:
             ds.info['sensor_types'] = sensor_types
             pipe = self._raw[self.get('raw')]
-            exclude = 'bads'
             for data_kind in sensor_types:
                 sysname = pipe.get_sysname(info, ds.info['subject'], data_kind)
                 adjacency = pipe.get_adjacency(data_kind)
@@ -2196,11 +2195,11 @@ class Pipeline(FileTree):
                 else:
                     name = data_kind
                 if variable_tmax:
-                    ys = [load.mne.epochs_ndvar(e, data=data_kind, sysname=sysname, adjacency=adjacency, exclude=exclude, name=data_kind)[0] for e in ds['epochs']]
+                    ys = [load.mne.epochs_ndvar(e, data=data_kind, sysname=sysname, adjacency=adjacency, name=data_kind)[0] for e in ds['epochs']]
                     if isinstance(data.sensor, str):
                         ys = [getattr(y, data.sensor)('sensor') for y in ys]
                 else:
-                    ys = load.mne.epochs_ndvar(ds['epochs'], data=data_kind, sysname=sysname, adjacency=adjacency, exclude=exclude)
+                    ys = load.mne.epochs_ndvar(ds['epochs'], data=data_kind, sysname=sysname, adjacency=adjacency)
                     if isinstance(data.sensor, str):
                         ys = getattr(ys, data.sensor)('sensor')
                 ds[name] = ys
@@ -2332,7 +2331,7 @@ class Pipeline(FileTree):
             sns_ndvar = 'both'
             del_epochs = False
         else:
-            raise ValueError(f'keep_epochs={keep_epochs!r}')
+            raise ValueError(f'{keep_epochs=}')
 
         ds = self.load_epochs(subject, baseline, sns_ndvar, reject=reject, cat=cat, samplingrate=samplingrate, decim=decim, pad=pad, data_raw=data_raw, vardef=vardef)
 
--- a/eelbrain/_experiment/mne_experiment.py
+++ b/eelbrain/_experiment/mne_experiment.py
@@ -244,7 +244,6 @@ class Pipeline(FileTree):
     # merge adjacent events in the stimulus channel
     merge_triggers: int = None
     # add this value to all trigger times (in seconds); global shift, or {subject: shift, (subject, session): shift} dictionary
-    # NOTE: Is this what we want?
     trigger_shift: Union[float, Dict[Union[str, Tuple], float]] = 0
 
     # variables for automatic labeling {name: {trigger: label, triggers: label}}
--- a/eelbrain/_experiment/mne_experiment.py
+++ b/eelbrain/_experiment/mne_experiment.py
@@ -62,7 +62,6 @@ from .test_def import (
     assemble_tests,
 )
 from .variable_def import GroupVar, Variables
-from . import preprocessing
 
 
 # current cache state version
@@ -76,13 +75,15 @@ CACHE_STATE_VERSION = 17
 #  15:  merge_triggers attribute, store in input_state
 #  16:  stim_channel attribute, store in input_state
 
-BIDS_ENTITY_KEYS = ['subject', 'session', 'task', 'acquisition', 'run', 'datatype', 'suffix', 'extension']
+BIDS_ENTITY_KEYS = ('subject', 'session', 'task', 'acquisition', 'run', 'split')
+BIDS_PATH_KEYS = ('datatype', 'suffix', 'extension', *BIDS_ENTITY_KEYS)
 BIDS_ENTITY_PREFIX_MAP = {
     'subject': 'sub',
     'session': 'ses',
     'task': 'task',
     'acquisition': 'acq',
     'run': 'run',
+    'split': 'split',
 }
 
 # paths
@@ -155,7 +156,7 @@ def guess_y(ds, default=None):
 def generate_bids_template(entities: set[str]) -> str:
     "Generate a BIDS filename template from entity names"
     parts = []
-    for name in ('subject', 'session', 'task', 'acquisition', 'run'):
+    for name in BIDS_ENTITY_KEYS:
         if name not in entities:
             continue
         prefix = BIDS_ENTITY_PREFIX_MAP[name]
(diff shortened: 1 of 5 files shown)


eelbrain/_experiment/tests/sample_experiment.py
```diff
@@ -36,7 +36,7 @@ class SampleExperiment(Pipeline):
     }
 
     raw = {
-        'tsss': RawMaxwell('raw', st_duration=10., ignore_ref=True, st_correlation=.9, st_only=True),
+        'tsss': RawMaxwell('raw', st_duration=10., ignore_ref=True, st_correlation=.9, st_only=True, st_overlap=False),
```
christianbrodbeck, line 39: Great catch @zhangwenchili ! @larsoner that looks like an undocumented API change in mne-Python (related to the change in number of good trials in the test you encountered)

eelbrain/_experiment/mne_experiment.py
```diff
@@ -243,6 +244,7 @@ class Pipeline(FileTree):
     # merge adjacent events in the stimulus channel
     merge_triggers: int = None
     # add this value to all trigger times (in seconds); global shift, or {subject: shift, (subject, session): shift} dictionary
+    # NOTE: Is this what we want?
```
christianbrodbeck: Can you clarify this question? Or we can discuss it on Tuesday.

Did you see the documentation at https://eelbrain.readthedocs.io/en/bids/experiment.html#eelbrain.pipeline.Pipeline.trigger_shift? (Where, by the way, could you update the subject names to BIDS subject names in the example?)

eelbrain/_experiment/mne_experiment.py
```diff
@@ -2126,7 +2066,7 @@ def load_epochs(
                     dss.append(ds)
             return combine(dss)
 
-        # TODO: add_bads can't be str by type annotation
+        # NOTE: add_bads can't be str by type annotation
```
christianbrodbeck: Can you remove this option? If it's just inside this method and does not make any tests fail then may be fine to do in this PR.

I don't remember why this is here and don't see why it would be needed. I think the proper thing to do when bads are needed but should not be rejected would be to call `.load_bads()` separately?

eelbrain/_experiment/mne_experiment.py
```diff
@@ -2521,11 +2462,8 @@ def load_events(
         elif data_raw:
             ds.info['raw'] = self.load_raw(add_bads, preload=self.preload)
 
-        ds.info['subject'] = subject
-        ds.info['session'] = session
-        ds.info['task'] = self.get('task')
-        ds.info['acquisition'] = self.get('acquisition')
-        ds.info['run'] = self.get('run')
+        for k, v in entities.items():
+            ds.info[k] = v
```
```suggestion
        ds.info.update(entities)
```
Does the same thing

eelbrain/_experiment/preprocessing.py
```diff
@@ -850,34 +758,30 @@ def load_ica(self, path: BIDSPath) -> mne.preprocessing.ICA:
     def _check_ica_channels(
             ica: mne.preprocessing.ICA,
             info: mne.Info,
-            raise_on_mismatch: bool = False,
-            raw_name: str = None,
-            subject: str = None,
             return_missing: bool = False,  # if ICA is missing channels, return those (they can be dropped in data)
     ) -> bool | tuple:
         "Check whether `ica` and `info` contain the same channels"
+        # NOTE: Does order matter here?
```
christianbrodbeck: It should not theoretically (I don't know practically), but channels should always be in the same order as they are in the raw file, so order mismatch without channel mismatch is very unlikely

eelbrain/_experiment/tests/sample_experiment.py
larsoner: See documentation here

https://mne.tools/dev/changes/v1.10.html#api-changes-by-deprecation

You should have been hitting `FutureWarning`s until 1.11 landed last month. Some changes in channel behavior would be expected from this.

I'd recommend adding a treat-warnings-as-errors line to your `conftest.py`, such a line would have turned the `FutureWarning` (which is otherwise too easy to ignore/miss in testing) into an error. In MNE-Python we treat all un-caught warnings [as errors](https://github.com/mne-tools/mne-python/blob/f4853ce70c16bcde2f558eead852d6c432581e9a/mne/conftest.py#L130-L214) in tests (other than ones we explicitly ignore), but I *think* you could be more targeted for example and just treat certain classes or warnings from certain modules (like `mne`) as errors if you wanted

christianbrodbeck: Thanks for the reminder @larsoner

gokuprasanna (commented):
I can load the raw files from the appleseed-bids dataset without any issues now. So, fix seems to have worked.

eelbrain/_experiment/preprocessing.py
```diff
@@ -944,31 +854,37 @@ def _make(
...
+        raw.info['bads'] = [ch for ch in self.load_bad_channels(path) if ch in raw.ch_names]
         ica = self.load_ica(path)
-        missing = self._check_ica_channels(ica, raw.info, raise_on_mismatch=True, raw_name=raw_name, subject=path.entities['subject'], return_missing=True)
+        missing = self._check_ica_channels(ica, raw.info, return_missing=True)
         if missing:
             raw.drop_channels(missing)
         ica.apply(raw)
         return raw
 
+    def _as_dict(self, args: Sequence[str] = ()) -> dict:
```
christianbrodbeck, line 874: Is there a reason to move this method (and several others below)? It makes the diff harder to read (and changes harder to trace in the future)

christianbrodbeck (changes requested):
Thanks @zhangwenchili ! See individual question/comments. Thanks @gokuprasanna for testing

eelbrain/_experiment/mne_experiment.py
zhangwenchili: I think I brought this up in one meeting before, when we decided to change the key to `(subject, session)`. I'm just wondering if it will differ between tasks/runs. From the documentation, it appears that this varies from dataset to dataset. So maybe keep it as it is for now, unless we encounter such a case in the future?

The documentation is correct here, because we're not using `sub-xxx` in state variables.

eelbrain/_experiment/mne_experiment.py
zhangwenchili: > Can you remove this option?

Do you mean remove this parameter or remove the string type?

eelbrain/_experiment/preprocessing.py
zhangwenchili: Not a significant reason, I just feel it's easier to find things by moving important methods to top. I can move it back :)

eelbrain/_experiment/mne_experiment.py
christianbrodbeck: I meant, can we remove the option for `add_bads` to be a `str`? Or are you aware of a situation where this option is used internally?

eelbrain/_experiment/preprocessing.py
zhangwenchili: So do you prefer set equal (without order) or list equal when `return_missing is False`?

eelbrain/_experiment/mne_experiment.py
zhangwenchili: Removed in https://github.com/Eelbrain/Eelbrain/pull/139/commits/bcf61785aba43db10d1f86a4fdb60ebb10953141

eelbrain/_experiment/mne_experiment.py
christianbrodbeck: Ah, yes, I think we could leave it for now. In the long run, I wonder whether we could delegate handling this to a sidecar file if it is M/EEG file-specific.

eelbrain/_experiment/preprocessing.py
christianbrodbeck: Ok for now. For the future, it makes reviewing harder because it adds more changed lines, so it would be better to do it in a separate style-only refactor

eelbrain/_experiment/preprocessing.py
christianbrodbeck: I'm not sure whether order difference would cause any problems down the line, so my hunch would be to keep list equal, but if order differences would not cause any issues then you're right that changing to set equal would make sense.

zhangwenchili: OK, let's keep the list equal then :)

eelbrain/_experiment/preprocessing.py
```diff
@@ -367,78 +272,61 @@ def _load(
...
             new_bads = sorted(set(old_bads).union(new_bads))
         # print change
         print(f"{old_bads} -> {new_bads}")
         if new_bads == old_bads:
             return
         # write new bad channels
-        mark_channels(path, ch_names=new_bads, status='bad')
+        if override:
+            mark_channels(path, ch_names='all', status='good', verbose=MNE_VERBOSITY)
+        if new_bads:  # mne_bids < 0.17 handles empty list as 'all'
```
christianbrodbeck: Would it be easier to just require `mne_bids >=0.17`?

christianbrodbeck (changes requested):
I pushed some small changes.

When I try to create ICA for one subject it looks like ~it's processing all subjects' data instead of just that one subject~. Do you have an idea why that is? I did not check whether that already occurred before this PR. If the reason for this is outside this PR then let's fix it separately to keep this PR manageable, just let me know. 

Image

[appleseed.py](https://github.com/user-attachments/files/24282962/appleseed.py)

EDIT (sorry I was not looking right):  It seems to be recomputing TSSS for the same subject repeatedly, rather than caching the result. Here is a screenshot of one I let run for longer. TSSS is cached, so you should see `INFO    :  Raw tsss: computing Maxwell...` exactly once.:

Screenshot 2025-12-22 at 8 21 44 AM

christianbrodbeck:
I updated my comment above

eelbrain/_experiment/preprocessing.py
zhangwenchili: https://github.com/Eelbrain/Eelbrain/pull/139/commits/c71953fd4c3068d7fc8b953dd90c1df5ec03dc3a

zhangwenchili:
Sorry for the mistake, there's duplicate `_make` in `CachedRawPipe`. Fixed in https://github.com/Eelbrain/Eelbrain/pull/139/commits/54a9ca3e0e70967f858461d178e748a24f061170.

Also, https://github.com/Eelbrain/Eelbrain/pull/139/commits/3deb40addfbbd554200a88e3ae4f056dee3b6162 seems not working because mne is not using `warning` module. I set `verbose='ERROR'`.

christianbrodbeck (approved):
Thanks @zhangwenchili !

I pushed another small commit for catching a filename warning, strangely this time the message is properly issued through the warnings module and can be ignored. This seems like an inconsistency in mne-Python. If you're interested you could address this with a PR to mne-Python to issue the first warning through the warnings module as well.

zhangwenchili:
Thanks for the catch. Happy new year!

christianbrodbeck:
Happy new year! :)

PR #147, opened 2026-01-23
Minor fixes

zhangwenchili:
Hi Dr. @christianbrodbeck , I have reviewed the code and tested again on MNE sample data. There are some GUI issues in `plot_` methods, but I guess it's fine since they are barely used now.

Another concern is that the operation [here](https://github.com/Eelbrain/Eelbrain/compare/bids...zhangwenchili:feat-cache?expand=1#diff-daf7dda682641b2596cba631902d9a9cc09b4bb1d1ae8059c4536d3010a3125cL805) is usually very slow for large datasets. Should we consider dropping the cache validation here?

### DIFF
--- a/eelbrain/_experiment/mne_experiment.py
+++ b/eelbrain/_experiment/mne_experiment.py
@@ -455,6 +455,7 @@ class Pipeline(FileTree):
             'raw-dir': join('{root}', 'sub-{subject}', 'ses-{session}' if self._sessions else '', '{datatype}'),
             'deriv-dir': join('{root}', 'derivatives'),
 
+            'raw-file': join('{raw-dir}', '{raw_basename}_{suffix}{extension}'),
             # one ica-file for each task group
             'ica-file': join('{deriv-dir}', 'ica', '{epoch_basename}_raw-{raw}_ica.fif'),  # hard-coded in RawPipe
             'trans-file': join('{deriv-dir}', 'trans', '{subject_session}_trans.fif'),
@@ -6792,7 +6793,8 @@ class Pipeline(FileTree):
                     ica = self.load_ica()
                     rows.append((subject, ica.n_components_, len(ica.exclude)))
                 except FileMissingError:
-                    if all(source_pipe.mtime(self._bids_path.copy().update(task=task), False) for task in pipe.task):
+                    path = self._bids_path.copy()
+                    if all(source_pipe.mtime(path.update(task=task), False) for task in pipe.task):
                         rows.append((subject, "No ICA-file", -1))
                     else:
                         rows.append((subject, "No data", -1))
@@ -6992,6 +6994,8 @@ class Pipeline(FileTree):
         mri_list = []
         mrisubject_list = []
         raw_list = []
+        datatype = self.get('datatype')
+        suffix = self.get('suffix')
         for subject in self.iter():
             subject_list.append(subject)
             mrisubject_ = self.get('mrisubject')
@@ -6999,8 +7003,8 @@ class Pipeline(FileTree):
             if raw:
                 query = BIDSPath(
                     subject=subject,
-                    datatype='meg',
-                    suffix='meg',
+                    datatype=datatype,
+                    suffix=suffix,
                     root=self.get('root'),
                 )
                 matches = query.match()
@@ -7026,7 +7030,7 @@ class Pipeline(FileTree):
         if mrisubject:
             ds['mrisubject'] = Factor(mrisubject_list)
         if raw:
-            ds['raw files'] = Factor(raw_list)
+            ds['raw_files'] = Factor(raw_list)
 
         if asds:
             return ds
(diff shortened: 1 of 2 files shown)


eelbrain/_experiment/mne_experiment.py
```diff
@@ -2812,7 +2813,7 @@ def load_evoked_stc(
 
         # preload morph matrices
         morph_sources = {subject.removeprefix('sub-') for subject in from_subjects.values() if subject != common_brain}
-        source_morphs = {subject: self.load_source_morph(subject=subject) for subject in morph_sources}
+        source_morphs = {'sub-' + subject: self.load_source_morph(subject=subject) for subject in morph_sources}
```
christianbrodbeck, line 2816: `from_subjects` is a {MEG-subject: MRI-subject} mapping, right? So `from_subjects.values()` may not all have the `sub-` prefix?

christianbrodbeck (changes requested):
Thanks @zhangwenchili !

One thing that *may* need changing (see comment)

Which operation is slow for large datasets? The link for me just shows the diff of this PR.

zhangwenchili:
Sorry for the wrong link, it should be [line 805](https://github.com/Eelbrain/Eelbrain/pull/147/files#diff-daf7dda682641b2596cba631902d9a9cc09b4bb1d1ae8059c4536d3010a3125cL805) in mne_experiment.py

eelbrain/_experiment/mne_experiment.py
zhangwenchili: Current implementation is: MEG subjects don't have the prefix to match the BIDSPath syntax, while MRI subjects have to align with the file/folder names in freesurfer directory.

christianbrodbeck: If I define custom MRI-subjects (overriding `Pipeline.mri_subjects`), isn't it possible that some of my MRI-subjects (i.e., items in `from_subjects.values()`) won't have the `sub-` prefix? Or is that always added internally?

I guess my own confusion indicates that it would be great to add an entry for `Pipeline.mri_subjects` to the [docs](https://eelbrain.readthedocs.io/en/bids/experiment.html#experiment-definition) and document how the names will behave? :)

christianbrodbeck:
> Another concern is that the operation [here](https://github.com/Eelbrain/Eelbrain/compare/bids...zhangwenchili:feat-cache?expand=1#diff-daf7dda682641b2596cba631902d9a9cc09b4bb1d1ae8059c4536d3010a3125cL805) is usually very slow for large datasets. Should we consider dropping the cache validation here?

Are you referring to

```py
events[key] = events_in = self.load_events(add_bads=False, data_raw=False)
```

If so: this *should* take long the first time the experiment is initialized, because extracting events requires reading the whole raw data file, but the events should then be cached and it should be much faster on subsequent initializations. Is that not so in your case?

eelbrain/_experiment/mne_experiment.py
zhangwenchili: Yes, currently the suffix is only added to the default mapping. Can we add the document entry to tell user to always add the suffix if they define their own mapping? The `mrisubject` state parameter must have prefix, and the mapping should be consistent with that.

zhangwenchili:
> > Another concern is that the operation [here](https://github.com/Eelbrain/Eelbrain/compare/bids...zhangwenchili:feat-cache?expand=1#diff-daf7dda682641b2596cba631902d9a9cc09b4bb1d1ae8059c4536d3010a3125cL805) is usually very slow for large datasets. Should we consider dropping the cache validation here?
> 
> Are you referring to
> 
> ```python
> events[key] = events_in = self.load_events(add_bads=False, data_raw=False)
> ```
> 
> If so: this _should_ take long the first time the experiment is initialized, because extracting events requires reading the whole raw data file, but the events should then be cached and it should be much faster on subsequent initializations. Is that not so in your case?

You're right. After testing, I found its `mne_bids.get_datatypes` taking much time. It takes ~130 seconds on ImageNet dataset. I will have a look at that function. If it's a behavior they expexted, we should avoid using this.

christianbrodbeck:
> After testing, I found its `mne_bids.get_field_values` taking much time. It takes ~130 seconds on ImageNet dataset. I will have a look at that function. If it's a behavior they expexted, we should avoid using this.

Thanks! Does it maybe need to read many sidecar files to be exhaustive?

christianbrodbeck:
Maybe we could cache the result, and only redo if any relevant file modification times change? (different PR though)

eelbrain/_experiment/mne_experiment.py
christianbrodbeck: > Can we add the document entry to tell user to always add the suffix if they define their own mapping? 

If it always needs the prefix, what about just automatically adding it if the user omits it?

christianbrodbeck: I think for the users, the most intuitive would be to just use the name of the subdirectory in `SUBJECTS_DIR`

zhangwenchili: > If it always needs the prefix, what about just automatically adding it if the user omits it?

Yeah this makes sense.

christianbrodbeck (approved):
Assuming MRI subject handling may be cleaner in it's own PR, feel free to merge this if ready

PR #149, opened 2026-01-27
fix Pipeline bad channel handling

christianbrodbeck:
- Pre-existing bad channels file that are missing a status column can be treated as all good
- Fix a bug where make_bad_channels([], redo=True) lead to marking all channels as bad

### DIFF
--- a/eelbrain/_experiment/mne_experiment.py
+++ b/eelbrain/_experiment/mne_experiment.py
@@ -4048,7 +4048,7 @@ class Pipeline(FileTree):
         merge_bad_channels : merge bad channel definitions for all tasks
         """
         pipe = self._raw[self.get('raw', **kwargs)]
-        pipe.make_bad_channels(self._bids_path, bad_chs, override=redo)
+        pipe.make_bad_channels(self._bids_path, bad_chs, redo=redo)
 
     def make_bad_channels_auto(self, flat=None, redo=False, **state):
         """Automatically detect bad channels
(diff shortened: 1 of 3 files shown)


zhangwenchili:
For the second one, maybe you have mne_bids < 0.17 because dev environment is not updated automatically? See https://github.com/Eelbrain/Eelbrain/pull/139#discussion_r2638035198.

christianbrodbeck:
You're right. But skipping the `mark_channels` call when there are no channels to mark should be good anyways right?

christianbrodbeck:
Also, looks like the issue would persist in mne-bids 0.18: https://github.com/mne-tools/mne-bids/blob/9139214b9377c2d079f9a45f93049ff3b6c7ae51/mne_bids/write.py#L2884 (copying permalink changed to a commit, but these lines are there in release 0.18 on Conda-forge, although not in main)

zhangwenchili:
> You're right. But skipping the `mark_channels` call when there are no channels to mark should be good anyways right?

Yes we can keep this. Other changes look good.


> Also, looks like the issue would persist in mne-bids 0.18

I just tried it works as they documented, maybe they used other ways to remove it.
image

zhangwenchili (approved):

christianbrodbeck:
> I just tried it works as they documented, maybe they used other ways to remove it.

Did you check the resulting channels file?

zhangwenchili:
Oh I need to have at least an existing bad channel to reproduce this, otherwise it will return early. :)

PR #150, opened 2026-01-27
fix RawReReference with reference='average'

christianbrodbeck:


eelbrain/_experiment/definitions.py
```diff
@@ -261,6 +261,28 @@ def typed_arg(arg, type_, secondary_type=None):
         return type_(arg)
```
zhangwenchili, line 263: `list_arg('test', 1)` will not trigger type error.

```
Traceback (most recent call last):
  File "/mnt/d/scripts/python.py", line 26, in <module>
    print(list_arg('test_tuple', 1))
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/d/scripts/python.py", line 21, in list_arg
    out = list(arg)
          ^^^^^^^^^
TypeError: 'int' object is not iterable
```

A possible solution:
```python3
def list_arg(
        name: str,  # for error message
        arg: Sequence,
        item_type: Type = str,
        allow_none: bool = True,
):
    match arg:
        case list():
            out = arg
        case tuple() | set():
            out = list(arg)
        case item_type():
            return [arg]
        case None:
            if allow_none:
                return None
            else:
                raise TypeError(f"{name}={arg!r}")
        case _:
            raise TypeError(f"{name}={arg!r}")

    if not all(isinstance(item, item_type) for item in out):
        raise TypeError(f"{name}={arg!r}: sequence of {item_type.__name__} required")
    
    return out
```

christianbrodbeck: Isn't `list_arg('test', 1)` --> `TypeError` the desired behavior?

zhangwenchili: Yes but now it will not trigger the one you defined, but another less informative one (see above).

christianbrodbeck: I refactored to have a better error message

zhangwenchili:
Thanks @christianbrodbeck ! Others look good.

zhangwenchili (approved):

PR #152, opened 2026-01-30
BIDS pipeline

christianbrodbeck:


christianbrodbeck:
@zhangwenchili anything you think we should address before merging?

@gokuprasanna this may have been the t-test test failure you mentioned, fixed in https://github.com/Eelbrain/Eelbrain/pull/152/commits/f65b89b57f4679520d498a09f0ac4337c9d33167

zhangwenchili:
Just one minor thing - It might be good to create a branch to keep a snapshot of old version. What about naming it `main-pre-bids`?

I had a look at https://github.com/Eelbrain/Eelbrain/pull/147#issuecomment-3797020165. The function is just walking through the folder names. I tried running the same function on a Windows environment and it's fast, so that is a WSL issue.

smiths:
@zhangwenchili, if you want to keep a snapshot of the entire repo, you can use tags.  [GitHub tags](https://medium.com/@shahriarhasan0_57376/what-are-github-tags-and-how-to-use-them-5e7abd4c2fd9) are permanent markers that point to specific points in a repository's history.  Would tags accomplish what you are looking for?

zhangwenchili:
The tags sounds like a better approach. I created one https://github.com/Eelbrain/Eelbrain/releases/tag/pre-bids-stable.

christianbrodbeck:
@zhangwenchili that reminded me of https://github.com/Eelbrain/Eelbrain/issues/153 -- Could you look into this at some point? Since it only affects custom MRI subject mappings it's probably not high priority, and we can do it after merging BIDS?

zhangwenchili:
Sure I will open a new PR after this merge.

christianbrodbeck:
> The function is just walking through the folder names. I tried running the same function on a Windows environment and it's fast, so that is a WSL issue.

@zhangwenchili maybe it would be worth mentioning under known issues? (https://eelbrain.readthedocs.io/en/stable/changes.html#known-issues)

zhangwenchili:
I think it's not the right place to put it since it has nothing to do with Eelbrain. It's just WSL's cross-OS file IO performance.

christianbrodbeck:
My concern is that someone experiences the same thing you experienced with the `Pipeline` and they're trying to find out what the issue is, we want them to be able to find out as efficiently as possible :) and maybe even warn them to not use WSL in advance. We can explain that it's not an issue in our control.

gokuprasanna:
> @zhangwenchili anything you think we should address before merging?
> 
> @gokuprasanna this may have been the t-test test failure you mentioned, fixed in [f65b89b](https://github.com/Eelbrain/Eelbrain/commit/f65b89b57f4679520d498a09f0ac4337c9d33167)

Thank @christianbrodbeck This resolved the issue I had when testing locally.

zhangwenchili:
> My concern is that someone experiences the same thing you experienced with the `Pipeline` and they're trying to find out what the issue is, we want them to be able to find out as efficiently as possible :) and maybe even warn them to not use WSL in advance. We can explain that it's not an issue in our control.

Added in https://github.com/Eelbrain/Eelbrain/pull/155

PR #155, opened 2026-02-02
Fix MRI suffix

### LINKED ISSUE #153: Automatic handling of the `sub-` prefix for MRI-subjects
Internally handle the `sub-` prefix, such that users can specify MRI-subject consistent with the name of the subdirectory in `SUBJECTS_DIR`.

Following discussion in https://github.com/Eelbrain/Eelbrain/pull/147#discussion_r2722730892

zhangwenchili:
Closes https://github.com/Eelbrain/Eelbrain/issues/153.
Fix parameter for slow test. Slow test itself works fine.
Add WSL cross-OS file access to know issues.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -9,6 +9,7 @@ Known issues
 
 Check for open issues, or report new ones on `GitHub <https://github.com/Eelbrain/Eelbrain/issues>`_.
 
+* If using WSL (Windows Subsystem for Linux), please store your dataset on the Linux filesystem (e.g., under ``/home/username/...``) instead of the Windows filesystem (e.g., ``/mnt/c/...``) to avoid its slow cross-OS file access.
 * Fixed in **0.38.3**, Windows only (`#52 <https://github.com/Eelbrain/Eelbrain/issues/52>`_): due to unexpected data loss in :class:`multiprocessing.sharedctypes.RawArray` for large arrays, permutation tests on large datasets using multiprocessing could return spurious results in which *p*-values for *all* clusters were reported as exactly 0.
 
 
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -12,7 +12,7 @@ Check for open issues, or report new ones on `GitHub <https://github.com/Eelbrai
 * Fixed in **0.38.3**, Windows only (`#52 <https://github.com/Eelbrain/Eelbrain/issues/52>`_): due to unexpected data loss in :class:`multiprocessing.sharedctypes.RawArray` for large arrays, permutation tests on large datasets using multiprocessing could return spurious results in which *p*-values for *all* clusters were reported as exactly 0.
 
 Issues on prior versions
----
+-----------
 
 * If using WSL (Windows Subsystem for Linux), please store your dataset on the Linux filesystem (e.g., under ``/home/username/...``) instead of the Windows filesystem (e.g., ``/mnt/c/...``) to avoid its slow cross-OS file access.
 
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -9,12 +9,13 @@ Known issues
 
 Check for open issues, or report new ones on `GitHub <https://github.com/Eelbrain/Eelbrain/issues>`_.
 
-* Fixed in **0.38.3**, Windows only (`#52 <https://github.com/Eelbrain/Eelbrain/issues/52>`_): due to unexpected data loss in :class:`multiprocessing.sharedctypes.RawArray` for large arrays, permutation tests on large datasets using multiprocessing could return spurious results in which *p*-values for *all* clusters were reported as exactly 0.
+* If using WSL (Windows Subsystem for Linux), please store your dataset on the Linux filesystem (e.g., under ``/home/username/...``) instead of the Windows filesystem (e.g., ``/mnt/c/...``) to avoid its slow cross-OS file access.
+
 
 Issues on prior versions
 -----------
 
-* If using WSL (Windows Subsystem for Linux), please store your dataset on the Linux filesystem (e.g., under ``/home/username/...``) instead of the Windows filesystem (e.g., ``/mnt/c/...``) to avoid its slow cross-OS file access.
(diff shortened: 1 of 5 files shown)


doc/changes.rst
```diff
@@ -9,6 +9,7 @@ Known issues
 
 Check for open issues, or report new ones on `GitHub <https://github.com/Eelbrain/Eelbrain/issues>`_.
 
+* If using WSL (Windows Subsystem for Linux), please store your dataset on the Linux filesystem (e.g., under ``/home/username/...``) instead of the Windows filesystem (e.g., ``/mnt/c/...``) to avoid its slow cross-OS file access.
 * Fixed in **0.38.3**, Windows only (`#52 <https://github.com/Eelbrain/Eelbrain/issues/52>`_): due to unexpected data loss in :class:`multiprocessing.sharedctypes.RawArray` for large arrays, permutation tests on large datasets using multiprocessing could return spurious results in which *p*-values for *all* clusters were reported as exactly 0.
```
christianbrodbeck: Would it make sense to insert a subheading like "Issues on prior versions"?

eelbrain/_experiment/mne_experiment.py
```diff
@@ -313,7 +313,7 @@ class Pipeline(FileTree):
     # MRI subject names: {subject: mrisubject} mappings
     # selected with e.set(mri=dict_name)
     # default is identity (mrisubject = subject)
-    mri_subjects = {'': keydefaultdict(lambda s: 'sub-' + s)}
+    mri_subjects = {'': keydefaultdict(lambda s: s)}
```
christianbrodbeck, line 316: What if a user *does* use `sub-` in `mri_subjects`? Am I right to assume that right now they would end up with `sub-sub-` labels? Will MRI SUBJECTs directories never have `sub-`?

Maybe you could update the corresponding documentation, to make clear what is expected?

christianbrodbeck (changes requested):
Thanks @zhangwenchili ! See suggestion/question in comments.

eelbrain/_experiment/mne_experiment.py
zhangwenchili: Yes, I should add a check before adding the suffix: https://github.com/Eelbrain/Eelbrain/pull/155/commits/17417a59719698e1ab81d579219127c587abdfa4

doc/changes.rst
zhangwenchili: I think that's kind of duplicated with "known issues" since known issues are also likely from previous versions.

christianbrodbeck: Right, hence I meant a sub-heading *under* known issues. My reasoning is that most people would be using an up to date version and may want to scan known issues in the current version when they encounter an issue.

eelbrain/_experiment/mne_experiment.py
christianbrodbeck: Is there a good reason to still treat the common brain as a special case with the new scheme?

zhangwenchili: I saw public dataset doing this, e.g. https://openneuro.org/datasets/ds000248/versions/1.2.4.

christianbrodbeck: Right, for `FreeSurfer` folders, but I meant the other way – since we're internally adding the `sub-` prefix for all subjects who don't have it in the `FreeSurfer` folders, we can just do the same of the common brain?

zhangwenchili: Do you mean adding the prefix for `fsaverage` folder? In this example they added prefix for subjects except fsaverage. I think there's a reason for it.

christianbrodbeck: No, I just mean it may not be needed as an explicit exception in the code

zhangwenchili: Do you mean line 1621 and 6448? For those lines, a suffix will be added to fsaverage if you do make an exception, which would be wrong if you want a folder name like https://openneuro.org/datasets/ds000248/versions/1.2.4.

christianbrodbeck: Ok, I think it _would_ be good to document this explicitly at https://eelbrain.readthedocs.io/en/latest/experiment.html#setting-up-the-file-structure

This does mention MRI subjects, but it does not mention `fsaverage`. We should mention the template brain explicitly.

Could you also document the `mrisubjects` attribute here? https://eelbrain.readthedocs.io/en/latest/experiment.html#experiment-definition

christianbrodbeck: P.S. we want to keep `fsaverage` as folder name, as it makes it easier to get the subject from FreeSurfer

doc/changes.rst
zhangwenchili: Added in https://github.com/Eelbrain/Eelbrain/pull/155/commits/7c4075eee8c05f21311b0716542ffddc1d526d59

christianbrodbeck: It should be switched :) the WSL issue is an issue that can be encountered in the current version, whereas the "Fixed in ..." issue is in a prior version

zhangwenchili: https://github.com/Eelbrain/Eelbrain/pull/155/commits/50e0c46bc419042b272c4cfa4cba320dc009b8f6

eelbrain/_experiment/mne_experiment.py
zhangwenchili: https://github.com/Eelbrain/Eelbrain/pull/155/commits/1b13751fbc6ce004577c6fa18bb7cbbc8e93c91d

doc/experiment.rst
christianbrodbeck, line 98: I think here it might be helpful to add fsaverage as an exception, right?

christianbrodbeck (changes requested):
Just one small suggestion on the docs

doc/experiment.rst
zhangwenchili: Added in https://github.com/Eelbrain/Eelbrain/pull/155/commits/1eb3dcf638a6d91924f51b48f4a733f0f58cf97a

christianbrodbeck:
Thanks @zhangwenchili !

PR #156, opened 2026-02-02
CI:  update codecov upload

### LINKED ISSUE #118: Investigate Codecov change coverage issue

christianbrodbeck:
@zhangwenchili I noticed there is now v5 (https://github.com/codecov/codecov-action)

re-added the secret on https://github.com/Eelbrain/Eelbrain

### DIFF
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -51,4 +51,3 @@ jobs:
       uses: codecov/codecov-action@v5
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
-        verbose: true  # optional (default = false)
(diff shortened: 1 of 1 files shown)


christianbrodbeck:
@zhangwenchili looks like this PR also uploaded, right? But if I understood https://github.com/Eelbrain/Eelbrain/issues/118#issuecomment-3836195563 correct that could be because it's from my fork?

zhangwenchili:
> @zhangwenchili looks like this PR also uploaded, right? But if I understood [#118 (comment)](https://github.com/Eelbrain/Eelbrain/issues/118#issuecomment-3836195563) correct that could be because it's from my fork?

Yes the new secret worked! In this PR it uploaded to Eelbrain/Eelbrain, which is why there are 6 successful checks below. :)

christianbrodbeck:
Great! So we don't really need the update in this PR? Maybe still a good idea?

zhangwenchili:
Not needed to address https://github.com/Eelbrain/Eelbrain/issues/118. But I think we can keep the v5 and turn off the verbosity?

PR #158, opened 2026-02-06
Fix emptyroom cov mtime

### LINKED ISSUE #157: Empty room covariance cache check
Issue reported [here](https://github.com/christianbrodbeck/TRF-Tools/issues/16)

@zhangwenchili could you have a look? It looks like the error happens when checking whether the covariance is up to date (not when creating it the first time). The cache check still seems to assume that "empty room" is a task, which it is not anymore.

zhangwenchili:
Closes https://github.com/Eelbrain/Eelbrain/issues/157


doc/experiment.rst
```diff
@@ -310,7 +310,7 @@ Empty room noise covariance
 
 To use empty room data for estimating the noise covariance, follow these steps:
 
- - Put an empty room recording in each subject's MEG directory, just like the other MEG files, with task name ``emptyroom``. If you want to use the same empty room file for all subjects you can make links instead of copies to save space.
+ - Put an empty room recording in each subject's MEG directory, just like the other MEG files, with task name ``noise``. If you want to use the same empty room file for all subjects you can make links instead of copies to save space.
```
christianbrodbeck: Shouldn't this just link to the documentation on how to add empty room data to a BIDS dataset?

eelbrain/_experiment/mne_experiment.py
```diff
@@ -1275,8 +1268,9 @@ def _cov_mtime(self):
                 self.set(epoch=cov.epoch)
                 return self._epochs_mtime()
             elif isinstance(cov, RawCovariance):
-                self.set(task=cov.session)
-                return self._raw_mtime()
+                empty_room_bids_path = self._bids_path.find_empty_room()
```
christianbrodbeck: This made me realize we have not migrated `RawCovariance` properly.

Originally, `RawCovariance` could use any task, not just empty room (although in practice I don't know if that ever makes sense). This also applies in `Pipeline.make_cov()`. 

Currently `RawCovariance` still has a `session` parameter but that is not used. Can you either remove the session parameter (I think that's fine until we think of a valid use case for it) or rename it to `task` and use it in the relevant places?

eelbrain/_experiment/preprocessing.py
```diff
@@ -698,8 +698,8 @@ class RawICA(CachedRawPipe):
     select ICA components that should be removed. The arguments to that function
     determine what data is used to visualize the component time courses.
     For example, to determine which components load strongly on empty room data,
-    use ``e.make_ica_selection(task='emptyroom')`` (assuming an
-    ``'emptyroom'`` task is present).
+    use ``e.make_ica_selection(task='noise')`` (assuming an
+    ``'noise'`` task is present).
```
christianbrodbeck: Could we possibly put the empty-room source file resolution into `load_raw` or `RawPipe`, so that load_raw with task='noise' would always return the correct empty room raw file?

christianbrodbeck (changes requested):

doc/experiment.rst
zhangwenchili: Updated in https://github.com/Eelbrain/Eelbrain/pull/158/commits/b1a01135518cc90b11bcc593f1417061649f2b08

eelbrain/_experiment/mne_experiment.py
zhangwenchili: I just removed in https://github.com/Eelbrain/Eelbrain/pull/158/commits/4c8f2283f6e0bb5163e570a4807bd618a42c3c93, it's easy to add back when there's a use case.

eelbrain/_experiment/preprocessing.py
zhangwenchili: Good idea! Do we want to use `BIDSPath.find_empty_room` to automatically find the empty room file or let user specify?

For API, I think we can use another way instead of exposing task='noise'. The empty room files are ignored by MNE-BIDS, and enabling it will cause inconsistency between the two ways of organizing empty room files. Currently the `emptyroom` subject and `noise` task are not in state parameters.

christianbrodbeck: > Good idea! Do we want to use BIDSPath.find_empty_room to automatically find the empty room file or let user specify?

Are there use cases that `BIDSPath.find_empty_room` would not cover? I think using `BIDSPath.find_empty_room` as starting point for now would be good, the more we can rely on standardized code the better.

> For API, I think we can use another way instead of exposing task='noise'. The empty room files are ignored by MNE-BIDS, and enabling it will cause inconsistency between the two ways of organizing empty room files. Currently the emptyroom subject and noise task are not in state parameters.

I thought you could "intercept" task='noise' and handle it differently from other tasks, so we don't have to add additional parameters. What API would you suggest?

zhangwenchili: I think it fits well with the idea of raw classes: same entities, different proccessing/files. We can add a `RawEmptyroom` and override `_load` and `mtime`.

christianbrodbeck: > I think it fits well with the idea of raw classes: same entities, different proccessing/files. We can add a `RawEmptyroom` and override `_load` and `mtime`.

But we want the empty-room data to be pre-processed the same way as task files, so they need to move through the same raw pipeline as the other task files, starting at `RawSource`.

zhangwenchili: OK I see. Abstractly it's a choice between treating the empty room files the same as other subjects/tasks and making them affiliated with regular files. If we need them to be pre-processed as well, the second option will not work.

christianbrodbeck: Yes, we need a noise estimate _after a given preprocessing regime_

zhangwenchili: > Could we possibly put the empty-room source file resolution into `load_raw` or `RawPipe`, so that load_raw with task='noise' would always return the correct empty room raw file?

I tried and realized this will not work. `find_empty_room` needs the original task to find the matching empty-room file. So if we want to load or make ica on an empty-room file, the only way is to include the `sub-emptyroom` and `task-noise` in state parameter and let user switch manually. Or we keep the second option, which is simply disallow processing an empty-room file.

christianbrodbeck: So is the problem that if we set `task='noise'`, we wouldn't know which task to use to locate that empty room file?

I do not think we want to make ICA on an empty-room file. I thought it could be informative to show how the ICA loads on the empty room file, but to be real, I've never actually done it. So it's not a priority.

Maybe `load_raw` and corresponding raw pipe function could have a single parameter that's `noise`, so when you say `load_raw(task=task, noise=True)`, it loads empty room file corresponding to `task`?

zhangwenchili: > So is the problem that if we set `task='noise'`, we wouldn't know which task to use to locate that empty room file?

Yes

> Maybe `load_raw` and corresponding raw pipe function could have a single parameter that's `noise`, so when you say `load_raw(task=task, noise=True)`, it loads empty room file corresponding to `task`?

Updated in https://github.com/Eelbrain/Eelbrain/pull/158/commits/1acc13e22cf8b85592cff2aaf16c9b2c1618062e

doc/experiment.rst
```diff
@@ -308,10 +308,7 @@ To reject trials based on a pre-determined threshold, a loop can be used::
 Empty room noise covariance
 ---------------------------
 
-To use empty room data for estimating the noise covariance, follow these steps:
-
- - Put an empty room recording in each subject's MEG directory, just like the other MEG files, with task name ``emptyroom``. If you want to use the same empty room file for all subjects you can make links instead of copies to save space.
- - Use the empty room covariance though :ref:`state-cov` with ``e.set(cov='emptyroom')``
```
christianbrodbeck, line 314: You still need the last line, set `cov='emptyroom'`, right?

eelbrain/_experiment/tests/test_sample_experiment.py
```diff
@@ -48,7 +48,7 @@ def test_sample():
 
     # covariance
     with e._temporary_state:
-        e.set(cov='emptyroom')
+        e.set(cov='emptyroom', raw='raw')
```
christianbrodbeck: Why set to `raw='raw'`?

christianbrodbeck: Actually, could we test two different `raw`, _and_ load the covariance twice, so that the cache check gets covered?

christianbrodbeck (changes requested):
Are there some feasible ways to increase coverage? Maybe in `test_sample()`, add some methods with `noise=True`?

eelbrain/_experiment/tests/test_sample_experiment.py
zhangwenchili: Sorry I thought maxwell can't be applied on empty-room recordings. It can, but needs separate raw definitions. Added tests for different raw in https://github.com/Eelbrain/Eelbrain/pull/158/commits/d25254cecaf5bfd016a9b304ad7db692210da04e.

doc/experiment.rst
zhangwenchili: Added back in https://github.com/Eelbrain/Eelbrain/pull/158/commits/d25254cecaf5bfd016a9b304ad7db692210da04e.

doc/experiment.rst
```diff
@@ -310,8 +310,9 @@ Empty room noise covariance
 
 To use empty room data for estimating the noise covariance, follow these steps:
 
- - Put an empty room recording in each subject's MEG directory, just like the other MEG files, with task name ``emptyroom``. If you want to use the same empty room file for all subjects you can make links instead of copies to save space.
- - Use the empty room covariance though :ref:`state-cov` with ``e.set(cov='emptyroom')``
+- Set up empty room data according to the `instruction in BIDS specification <https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetoencephalography.html#empty-room-meg-recordings>`_.
+
```
christianbrodbeck: Let's stay consistent with layout (no empty line within bullet list)

eelbrain/_experiment/tests/test_sample_experiment.py
christianbrodbeck: Why does it need a separate raw definition?

christianbrodbeck (changes requested):

doc/experiment.rst
zhangwenchili: Updated in https://github.com/Eelbrain/Eelbrain/pull/158/commits/0efc8ed6338cddfe5fe8706033a1fb6cdba6fbcb

eelbrain/_experiment/tests/test_sample_experiment.py
zhangwenchili: Because it needs `coord_frame='meg'` (https://mne.tools/stable/generated/mne.preprocessing.maxwell_filter.html).

christianbrodbeck: Could we check inside `RawMaxwell` whether `dev_head_t` is available and modify the keyword dynamically?

zhangwenchili: Added in https://github.com/Eelbrain/Eelbrain/pull/158/commits/e7f7756c528f1d1fe679803d13610911c4d30a88

eelbrain/_experiment/tests/test_sample_experiment.py
```diff
@@ -48,9 +49,14 @@ def test_sample():
...
     with e._temporary_state:
-        e.set(cov='emptyroom')
+        e.set(cov='emptyroom', raw='tsss')
+        cov = e.load_cov()
+        assert isinstance(cov, mne.Covariance)
+        assert e.load_bad_channels(noise=True) == []
+        e.set(cov='emptyroom', raw='1-40')
         cov = e.load_cov()
         assert isinstance(cov, mne.Covariance)
+        assert e.load_bad_channels(noise=True) == []
```
christianbrodbeck, line 59: Could we add a second `load_cov` call with identical parameters to test that the cache check works?

zhangwenchili: Now we have two `load_cov` calls with `raw='tsss'` and `raw='1-40'`. What do you mean by "the cache check works"? Do you mean checking whether an identical call will use the cache instead of making a new one?

christianbrodbeck: I mean two calls with raw='1-40'. The first time the file does not exist so it will be generated. The second time the file exists and the pipeline has to check the file modification time. So these two calls would be testing different lines in the code.

zhangwenchili: Added in https://github.com/Eelbrain/Eelbrain/pull/158/commits/73943d0b6596941dc39f2bfbb1af81a8d055be57

christianbrodbeck:
Thanks @zhangwenchili ! @nigelflower can you test this branch or do you prefer me to make a release for testing?

nigelflower:
It seems to have fixed the error. I will test further and make sure that it is functioning properly, but I am optimistic.

nigelflower:
So it seems that there are some more issues. @zhangwenchili 's fix correctly retrieves the empty_room_bids_path, but then when the code tries to get the raw file, then there is an error.

In mne_experiment.py on line 4233, when we load the raw file, the info object does not seem to have a value for the key 'dig'

Screenshot 2026-02-10 at 11 15 09 AM

Screenshot 2026-02-10 at 11 16 55 AM

nigelflower:
And here is the very end of the stack trace for the error

Screenshot 2026-02-10 at 11 17 40 AM

christianbrodbeck:
Could it make sense to add a set of tests to Eelbrain based on (a subset of) the Appleseed data?

@zhangwenchili let me know if access to a Mac Studio would help you for testing with such a larger dataset.

zhangwenchili:
@christianbrodbeck I'll try again to test on my computer and let you know if the memory issue persists. Nigel said there's a default experiment file in the Appleseed repository. If that file has some parameters defined for this dataset, maybe I can start with that?

christianbrodbeck:
Work with Nigel for the proper pipeline subclass

zhangwenchili:
@christianbrodbeck I do need a better machine to test stc methods or multiple subjects on this dataset.

For the above error, it's because maxwell filter requires both `info["dev_head_t"]` and `info["dig"]` if `coord_frame='head'`.

Empty room files in this dataset has `info["dev_head_t"]` but no `info["dig"]`. But when I try to use `coord_frame='meg'`, it will cause `RuntimeError: Matrix is badly conditioned: 2184 >= 1000`.

EDIT: Error happened when trying to get a decomposition matrix and pseudoinverse matrices. Looks like its not caused by the dev_head_t matrix but some other parts in raw data.

nigelflower:
Hmm, I'm still getting the error you mentioned. `RuntimeError: Matrix is badly conditioned: 2184 >= 1000`

zhangwenchili:
> Hmm, I'm still getting the error you mentioned. `RuntimeError: Matrix is badly conditioned: 2184 >= 1000`

Yes I mean this looks like a problem in the data?

zhangwenchili:
UPDATE: I believe there's something wrong in the data, but I couldn't find it out. I can reproduce the same error without using the pipeline. There's a same error [here](https://mne.discourse.group/t/runtimeerror-matrix-is-badly-conditioned/9177), but that user passed the origin with wrong unit, which doesn't fit in our case.

zhangwenchili:
@christianbrodbeck The "Matrix is badly conditioned" can be solved by `make_bad_channels_auto` (it will mark 'MEG 056'). I removed automatically marking bad channels in `load_bad_channels` before, because I thought it's not good to apply a hard-coded value automatically. Now it looks like its necessary in such cases, and I'm adding it back.

Also there are another two questions

- You said we don't want to do ICA on empty-room recordings, but currently when you set `raw='ica'` and do `load_cov` it will  use the RawICA. Is this use case useful or we just raise an error?

- The Appleseed experiment file has `tmax='length'` in PrimaryEpoch, but by definition it should be float. And it will cause error when applying `Dataset.eval` on `tmax`.

christianbrodbeck:
@zhangwenchili so is it because 056 is flat? I think that's the only mind of bad channel that is always excluded in `make_bad_channels_auto` (and reasonably so because there is no data left in those :) ).

We do want to subtract the ICA from the empty room data as well – we want the processing to be as parallel as possible to the task file, that will give us the most realistic estimate of the noise in the task file!

It's an omission in the docs – `tmax` can be a string referring to a column in the events dataset – in the Appleseed case, the `length` column is added in `Appleseed.label_events`

zhangwenchili:
OK I see. But to compute ICA for empty-room, user needs to define a RawICA that includes task=noise, or we always combine empty-room with the corresponding task data?

christianbrodbeck:
Actually, you _want_ to subtract the task based ICA components. If you're projecting something out of the task data, it won't cause any noise in the task data, so you also want to project it out of your noise estimate.

zhangwenchili:
> Actually, you _want_ to subtract the task based ICA components. If you're projecting something out of the task data, it won't cause any noise in the task data, so you also want to project it out of your noise estimate.


Emm, so do you mean _compute_ the ICA  only from the task data and _apply_ to both?

christianbrodbeck:
> Emm, so do you mean _compute_ the ICA only from the task data and _apply_ to both?

Exactly. Currently the RawICA pipe computes exactly one decomposition per subject/session (even if there are multiple tasks, unless you already addressed that :) ). So it should be straight forward to associate one empty room noise covariance per subject/session?

zhangwenchili:
I think it will compute the ICA based on the tasks you specified in RawICA. The pipeline state will always point to one raw file, we can apply the ICA computed from that raw file to emptyroom data.

christianbrodbeck:
> I think it will compute the ICA based on the tasks you specified in RawICA. The pipeline state will always point to one raw file, we can apply the ICA computed from that raw file to emptyroom data.

Yes to both, that sounds right

zhangwenchili:
@christianbrodbeck Since we don't always create the channels.tsv now, I put the `make_bad_channels_auto` in `__init__` instead of when creating the channels.tsv.

It's hard to do the ICA part for now, since raw classes don't have access to the pipeline state, so they can't find the ICA file from task recording. It's better to do it after the refactor.

christianbrodbeck:
> It's hard to do the ICA part for now, since raw classes don't have access to the pipeline state, so they can't find the ICA file from task recording. It's better to do it after the refactor.

Do you have a suggestion for how to solve that? Should the `noise: bool = False` parameter actually be resolved in `RawPipe.load()` (and its subclasses)?

zhangwenchili:
My initial thinking is, after the refactor raw classes will have access to a `FileTree` instance and state parameters, which will solve this in an easier way. However, it looks like new issues and use cases keep occurring in the testing, maybe we can prioritize making the pipeline more functionally stable on Appleseed (by testing the full data on the server and testing tone files on CI)?


> Should the noise: bool = False parameter actually be resolved in RawPipe.load() (and its subclasses)?

This looks like the best way if we want to solve the issue without refactoring.

christianbrodbeck:
> > Should the noise: bool = False parameter actually be resolved in RawPipe.load() (and its subclasses)?
> 
> This looks like the best way if we want to solve the issue without refactoring.

I'd say, let's do that for now, so people can start working with the dataset, and then if you find a more elegant way to refactor later, even better?

zhangwenchili:
ICA for empty room data works now. I'm trying to run source localization. The server disk is quite full now, I may need some extra disk space (5~10G) to download and unzip the full dataset (second version Prasanna uploaded).

zhangwenchili:
Tested `load_epochs_stc` and `load_evoked_stc` with 

```
    defaults = {
        'raw': 'ica',
        'rej': '',
        'cov': 'emptyroom',
        'inv': 'fixed-6-MNE-0',
    }
```

for three subjects, and it works fine. I have asked Prasanna for the "appleseed_stimuli.txt" file and the mapping from original subject names to the renamed ones. With those I will be able to test with other epoch definitions.

christianbrodbeck:
Great! I've added you to https://github.com/christianbrodbeck/Appleseed which is the legacy version of the `Appleseed` subclass, but has `appleseed_stimuli.txt`. @nigelflower are you able to verify this works for your purpose now? @zhangwenchili do you want to add the preliminary test to this PR? (For now, you could mark it to only run when the data is available.) Or do you prefer to do that separately?

zhangwenchili:
> do you want to add the preliminary test to this PR? (For now, you could mark it to only run when the data is available.) Or do you prefer to do that separately?

I'll add the tests using a separate PR after testing other epoch definitions locally :)

nigelflower:
I am testing out the new commits, but there now seems to be an error with loading in the Raw files. It seems that the ica file (which seems to be the raw data) contains the bad channels that get removed in the emptyroom recording. Thus, this check fails.

Screenshot 2026-02-17 at 10 11 15 AM

That being said, the problem with the Maxwell filtering does seem to have been resolved!

nigelflower:
So, I think the problem is that when we read in the raw file to undergo ICA, we check to see if there are any channels in the raw file that are labeled as bad in the emptyroom. Since channel MEG 056 is labeled as bad in the emptyroom, then we return an error.

Screenshot 2026-02-17 at 10 38 53 AM

However, we do add MEG 056 as a bad channel later on here:

Screenshot 2026-02-17 at 10 40 39 AM

But this step seems to apply later than when we're making the comparison between the raw file and the empty room file.

christianbrodbeck:
Yes, what we would want is simply take the union of bad channels.

However, this *is* a case where you may want to consider using a different empty room file to avoid losing a whole channel in the task data. So it would be good making the user aware in some way that a channel will be discarded just because of the empty room data.

zhangwenchili:
I tried `load_raw` and didn't see the error. Did you produce ICA before this change? If yes, deleting the ica folder (`{root}/derivatives/ica`) and eelbrain cache (`{root}/derivatives/eelbrain`) may solve this.

nigelflower:
Ok, deleting the cache resolved this issue. Furthermore, I was able to load in the covariance matrix without any problems. I think everything works!

eelbrain/_experiment/mne_experiment.py
```diff
@@ -802,6 +796,10 @@ def __init__(
                     if self.check_raw_mtime:
                         log.debug("Raw file missing: %s", self._bids_path.fpath)
                     continue
+                # bad channels
+                self.make_bad_channels_auto()
+                if self._bids_path.datatype == 'meg' and self._bids_path.find_empty_room().fpath.exists():
+                    self.make_bad_channels_auto(noise=True)
```
christianbrodbeck: If multiple raw files share an empty room recording, this is going to process those files multiple times, right?

Another concern, if channels are already marked as good or bad, we might not want to run this.

Maybe we can address both these issues by checking whether the sidecar file already has a channel good/bad column, and only run auto rejection if it does not?

eelbrain/_experiment/mne_experiment.py
```diff
@@ -1879,21 +1877,24 @@ def load_annot(self, **state):
...
         return mne.read_labels_from_annot(self.get('mrisubject'), self.get('parc'), 'both', subjects_dir=self.get('mri-sdir'))
 
-    def load_bad_channels(self, **kwargs):
+    def load_bad_channels(self, noise: bool = False, **kwargs):
         """Load bad channels
 
         Parameters
         ----------
+        noise
+            Load bad channels for noise empty-room recording instead of the subject recording.
```
```suggestion
            Load bad channels for empty-room noise recording instead of the subject recording.
```

eelbrain/_experiment/mne_experiment.py
```diff
@@ -3301,13 +3303,16 @@ def load_raw(
             the ``tstart`` will be set to ``t = 0``.
         tstop
             Crop the raw data.
+        noise
+            Load corresponding empty-room data instead of current subject's raw data (default ``False``).
```
```suggestion
            Load corresponding empty-room data instead of current subject's task data (default ``False``).
```

eelbrain/_experiment/mne_experiment.py
```diff
@@ -4055,6 +4060,8 @@ def make_bad_channels(self, bad_chs=(), redo=False, **kwargs):
             in the raw data, a ValueError is raised.
         redo : bool
             If the file already exists, replace it (instead of adding).
+        noise : bool
```
christianbrodbeck: Please use type hints instead of docstring annotation whenever possible (including updating the other parameters :) )

eelbrain/_experiment/mne_experiment.py
```diff
@@ -4079,13 +4087,16 @@ def make_bad_channels_auto(self, flat=None, redo=False, **state):
             are considered bad (default 1e-14 for MEG and 0 for EEG).
         redo : bool
             If the file already exists, replace it (instead of adding).
+        noise : bool
```
christianbrodbeck: As above, please use type hints

eelbrain/_experiment/preprocessing.py
```diff
@@ -329,7 +338,9 @@ def make_bad_channels_auto(
             path: BIDSPath,
             flat: float = None,
             redo: bool = False,
+            noise: bool = False,
     ) -> None:
+        path = path if not noise else path.find_empty_room()
```
```suggestion
        if noise:
            path = path.find_empty_room()
```

christianbrodbeck: More readable?

eelbrain/_experiment/preprocessing.py
```diff
@@ -387,7 +398,9 @@ def mtime(
             self,
             path: BIDSPath,
             bad_chs: bool = True,
+            noise: bool = False,
     ) -> float:
+        path = path if not noise else path.find_empty_room()
```
```suggestion
        if noise:
            path = path.find_empty_room()
```

eelbrain/_experiment/preprocessing.py
```diff
@@ -969,24 +990,26 @@ def _make(
...
+        mtime = CachedRawPipe.mtime(self, path, bad_chs, noise=noise)
         if mtime:
             ica_mtime = self.ica_source.mtime(path, bad_chs)
             if ica_mtime:
                 return max(mtime, ica_mtime)
 
 
 class RawMaxwell(CachedRawPipe):
-    """Maxwell filter raw pipe
+    """Maxwell filter raw pipe. For empty room recordings, there is no `dev_head_t` information, `coord_frame = 'meg'` will be used automatically.
```
christianbrodbeck: I would put this under a "Notes" heading (first line should be a short summary, https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard)

christianbrodbeck (changes requested):
Just some minor things, thanks @zhangwenchili !

eelbrain/_experiment/preprocessing.py
zhangwenchili: Updated in https://github.com/Eelbrain/Eelbrain/pull/158/commits/9ea5379f2520c2e1b6ec20ef7a71477e924938ce

eelbrain/_experiment/preprocessing.py
zhangwenchili: https://github.com/Eelbrain/Eelbrain/pull/158/commits/9ea5379f2520c2e1b6ec20ef7a71477e924938ce

eelbrain/_experiment/preprocessing.py
zhangwenchili: https://github.com/Eelbrain/Eelbrain/pull/158/commits/9ea5379f2520c2e1b6ec20ef7a71477e924938ce

eelbrain/_experiment/mne_experiment.py
zhangwenchili: https://github.com/Eelbrain/Eelbrain/pull/158/commits/9ea5379f2520c2e1b6ec20ef7a71477e924938ce

eelbrain/_experiment/mne_experiment.py
zhangwenchili: https://github.com/Eelbrain/Eelbrain/pull/158/commits/9ea5379f2520c2e1b6ec20ef7a71477e924938ce

eelbrain/_experiment/mne_experiment.py
zhangwenchili: https://github.com/Eelbrain/Eelbrain/pull/158/commits/9ea5379f2520c2e1b6ec20ef7a71477e924938ce

eelbrain/_experiment/mne_experiment.py
zhangwenchili: https://github.com/Eelbrain/Eelbrain/pull/158/commits/9ea5379f2520c2e1b6ec20ef7a71477e924938ce

eelbrain/_experiment/mne_experiment.py
zhangwenchili: I think we can use the existence of input_state.pickle to do this: https://github.com/Eelbrain/Eelbrain/pull/158/commits/b3bc19a91776bb2935c21df85a173f8f519fbf03

eelbrain/_experiment/preprocessing.py
```diff
@@ -1027,6 +1029,10 @@ class RawMaxwell(CachedRawPipe):
     See Also
     --------
     Pipeline.raw
+
+    Notes
+    -----
+    For empty room recordings, there is no `dev_head_t` information, `coord_frame = 'meg'` will be used automatically.
```
christianbrodbeck: In RST, code requires double ` (check output of the build at https://eelbrain--158.org.readthedocs.build/en/158/generated/eelbrain.pipeline.RawMaxwell.html#eelbrain.pipeline.RawMaxwell)

eelbrain/_experiment/mne_experiment.py
christianbrodbeck: I don't think that works all the time. The user might start an analysis with a subset of subjects and then add more subjects later on.

christianbrodbeck (changes requested):

eelbrain/_experiment/mne_experiment.py
zhangwenchili: > checking whether the sidecar file already has a channel good/bad column

This will not always work either. Like in Appleseed, the intial states are all good but MEG 056 is not usable.

What about storing it in input_state? https://github.com/Eelbrain/Eelbrain/pull/158/commits/fe4efc4657e869756162a3cbdd756529545b249f

christianbrodbeck: > What about storing it in input_state?

Yeah I guess that's our last resort :)

Could we do it as part of `RawMaxwell`'s `make` method?

zhangwenchili: > Could we do it as part of RawMaxwell's make method?

We can, but I think you want the bad channels to be consistent across all preprocessing steps, right?

christianbrodbeck: Actually, I don't think that enforcing this is desirable. Theoretically the right kind of data cleaning could turn a bad channel into a good channel.

zhangwenchili: Emm, I didn't get the idea. We can add an attribute to enable/disable this check, if enforcing is not desirable. And if you can do data cleaning, that isn't really a counterexample to this, because that happens before the pipeline and applies to all preprocessing steps?

christianbrodbeck: 1) Right now, the pipeline assumes that there is just one bad channel definition. 

2) Ideally, bad channels could differ between raw stages. For example, assume a channel that has a lot of high frequency noise 30 Hz+. That channel should be marked bad if you analyze the unfiltered signal. However, if you low pass filter at 20 Hz you can remove that high frequency noise, so after the filter the channel should be marked as good.

3) I think it would be cleanest to keep the original BIDS channels file untouched whenever possible.

The pipeline does not have the infrastructure for 2 currently, so that would be a bigger project. but I think it should be okay to add bad channels into the raw file during RawMaxwell, and it would be conceptually consistent with 2 and 3 for the future.

zhangwenchili: Oh I see, I thought 2 was undesired. Thanks for the explanation!

zhangwenchili: https://github.com/Eelbrain/Eelbrain/pull/158/commits/f513c97f5098332e3122230f5ab2141a524f42af

eelbrain/_experiment/preprocessing.py
```diff
@@ -329,22 +338,28 @@ def make_bad_channels_auto(
...
-            if path.datatype == 'meg':
-                flat = 1e-14
-            elif path.datatype == 'eeg':
-                return
-            else:
-                raise NotImplementedError(f"{path.datatype=}")
-        elif flat == 0:
+        if noise:
+            path = path.find_empty_room()
+        if path.datatype == 'eeg' or flat == 0:
```
christianbrodbeck: This is not quite equivalent -- for EEG, the *default* should be to skip, but it should still be possible to *explicitly* mark flat channels as bad.

eelbrain/_experiment/preprocessing.py
```diff
@@ -1017,11 +1056,14 @@ def _make(
             self,
             path: BIDSPath,
             preload: bool,
+            noise: bool = False,
     ) -> mne.io.BaseRaw:
-        raw = self.source.load(path)
-        self.log.info("Raw %s: computing Maxwell filter for %s", self.name, path.fpath)
+        flat_chs = self._detect_flat_channels(path)
+        raw = self.source.load(path, add_bads=flat_chs, noise=noise)
```
christianbrodbeck: This is slightly inefficient because it needs to read the raw data from disk twice. Can we just detect bad channels on the `raw` here _after_ loading it? You can just append to `raw.info['bads']`.

Technically, we need to make sure we exclude flat channels right here, and not at the `RawSource`. Granted, I can't think of a realistic scenario in which a preprocessing step would make a non-flat channel flat.

christianbrodbeck (changes requested):
Getting close, some details in the latest iteration :)

eelbrain/_experiment/preprocessing.py
zhangwenchili: Changed in https://github.com/Eelbrain/Eelbrain/pull/158/commits/d9a2a60ee80745f8a29e1a4bf754ea534fcfe962. Should we make an exception for EEG here as well?

eelbrain/_experiment/preprocessing.py
```diff
@@ -1025,13 +1015,13 @@ class RawMaxwell(CachedRawPipe):
...
         Name of the raw pipe to use for input data.
-    bad_condition : str
+    bad_condition
         How to deal with ill-conditioned SSS matrices; by default, an error is
         raised, which might prevent the process to complete for some subjects.
         Set to ``'warning'`` to proceed anyways.
-    cache : bool
+    cache
         Cache the resulting raw files (default ``True``).
     ...
```
christianbrodbeck, line 1028: Add `flat`?

eelbrain/_experiment/preprocessing.py
```diff
@@ -1043,23 +1033,35 @@ class RawMaxwell(CachedRawPipe):
...
             path: BIDSPath,
             preload: bool,
             noise: bool = False,
     ) -> mne.io.BaseRaw:
-        flat_chs = self._detect_flat_channels(path)
-        raw = self.source.load(path, add_bads=flat_chs, noise=noise)
+        raw = self.source.load(path, noise=noise)
+        sysname = self.get_sysname(raw.info, path.subject, None)
+        adjacency = self.get_adjacency(None)
+        raw_ndvar = load.mne.raw_ndvar(raw, sysname=sysname, adjacency=adjacency)
```
christianbrodbeck: `None` in `self.get_sysname` and `self.get_adjacency` are marked as type mismatch (should be `str`). Should this be `path.datatype`?

christianbrodbeck (changes requested):

eelbrain/_experiment/preprocessing.py
zhangwenchili: Updated in https://github.com/Eelbrain/Eelbrain/pull/158/commits/90189916ccb8b6475ceda64bc0397d1be861bfba

eelbrain/_experiment/preprocessing.py
zhangwenchili: Updated in https://github.com/Eelbrain/Eelbrain/pull/158/commits/90189916ccb8b6475ceda64bc0397d1be861bfba

eelbrain/_experiment/mne_experiment.py
```diff
@@ -756,6 +750,7 @@ def __init__(
         # ==============================
         events = {}  # {(subject, recording): event_dataset}
         self._stim_channel = sequence_arg(f'{self.__class__.__name__}.stim_channel', self.stim_channel)
+        init_bad_channels = False
```
christianbrodbeck: This is now redundant

christianbrodbeck (changes requested):

eelbrain/_experiment/mne_experiment.py
zhangwenchili: https://github.com/Eelbrain/Eelbrain/pull/158/commits/f6c04f4f2f935600d533ba43b5df0a229e9fc5a1

christianbrodbeck:
Thanks @zhangwenchili !

PR #162, opened 2026-02-25
chore: modernize code with pre-commit hooks

christianbrodbeck:
Most changes are from automatic tools. Pre-commit is now used for CI and can be used locally. See https://github.com/Eelbrain/Eelbrain/commit/0fa7bf5507d4735a1be8ec95dbe2b9bee177bb39

Want to get this in while we don't have any major PRs open.


christianbrodbeck:
@zhangwenchili @Daffodil404 , please check out changes to `doc/development.rst` (and possibly `.pre-commit-config.yaml`, `.github/workflows/code-quality.yml`). If these make sense, please let me know/approve the PR.

eelbrain/_experiment/mne_experiment.py
```diff
@@ -104,7 +105,7 @@
 
 
 # Argument types
-BaselineArg = Union[bool, Tuple[Optional[float], Optional[float]]]
+BaselineArg = Union[bool, tuple[Optional[float], Optional[float]]]
```
zhangwenchili: Remove the union here as well? (and next three lines)
```suggestion
BaselineArg = bool | tuple[float | None, float | None]
```

zhangwenchili (commented):
Looks good! Just a small suggestion.

eelbrain/_experiment/mne_experiment.py
christianbrodbeck: Looks like `pyupgrade` misses types outside function signatures, fixed in https://github.com/Eelbrain/Eelbrain/pull/162/commits/be82a4d56c4b74789473f5cec22bd925623dd6d3 (manually,  not as automatic check unfortunately)

env-dev.yml
```diff
@@ -45,11 +45,11 @@ dependencies:
 - sphinxcontrib-bibtex
 - pydocstyle
 # testing
-- flake8
+- pre-commit
 - pytest
 - pytest-cov
 - pytest-faulthandler
-- pingouin
+- pingouin >= 0.6.0  # breaking API used in tests
```
Daffodil404, line 52: Is the version for `pingouin` need to be detailed set here? Since I checked that 0.6.0 is the latest version and it will always get the latest version by default.

Daffodil404:
I left a comment about the automatic dependency version control, everything is fine.

Daffodil404 (approved):

env-dev.yml
christianbrodbeck: It's safer to make it explicit. `mamba` using the latest version is not guaranteed unless you pin it.

PR #175, opened 2026-04-30
Fix repeat

huzafahyde:
Fixed NDVar.repeat(), previously wouldn't work if NDVar already had 'case' dimension because it tried to make a new NDVar with the old case, which had a mismatch to the # of sides

### DIFF
--- a/eelbrain/_ndvar/tests/test_ndvar.py
+++ b/eelbrain/_ndvar/tests/test_ndvar.py
@@ -135,6 +135,7 @@ def test_diff():
     assert_array_equal(diff.x.data, np.diff(y.x, 1, 1, y.x[:, :1]))
     assert_array_equal(diff.x.mask, target_mask)
 
+
 def test_repeat():
     ds = datasets.simulate_erp()
     y = ds['eeg']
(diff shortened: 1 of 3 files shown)


christianbrodbeck:
Can you rebase on `main` and force push? That should fix the CI environments.

eelbrain/_ndvar/tests/test_ndvar.py
```diff
@@ -136,6 +136,17 @@ def test_diff():
     assert_array_equal(diff.x.mask, target_mask)
 
 
+def test_repeat():
```
christianbrodbeck: Could you move this test to `eelbrain.tests.test_data` and maybe name is `test_ndvar_methods` so we can add other method tests in the future? `eelbrain/_ndvar/tests/test_ndvar` actually contains tests of `NDVar` *operations* defined in `eelbrain/_ndvar`.

christianbrodbeck (changes requested):

christianbrodbeck:
@huzafahyde any chance you could finish this in the near future?

christianbrodbeck (approved):
Thanks @huzafahyde

PR #177, opened 2026-05-15
Clean up docs

larsoner:
While working on https://github.com/Eelbrain/Eelbrain/pull/133 I noticed several warnings during doc build. That PR fixes a few of them (in files I was already touching), this fixes the rest. I expect it to fail until https://github.com/Eelbrain/Eelbrain/pull/133 lands and I can merge `main` into this branch, though. Hopefully it does, actually, it would indicate the `fail_on_warning: true` RTD config value works correctly!

1. Build with `-WT --keep-going` which means "print errors, treat them as warnings, but don't stop on the first error". This prevents many little linking bugs from popping up during builds.
2. Modernize `sphinx-gallery` conf to allow it to be pickled (this was a warning, and prevents proper build caching for incremental doc updates / builds)
3. Fix the other errors (they are hopefully sufficiently self-explanatory looking at the `diff` but can explain them inline if not)

In many repos I actually build with `-n` as well, which means "nitpicky" where it gets even more particular about finding things like `:py:obj:` link errors and the like. I tried enabling it here but there were way too many warnings, so I'm leaving that as a follow-up for someone if they're motivated.

*EDIT:*

Buliding locally I get:
```
/home/larsoner/python/Eelbrain/doc/development.rst:157: WARNING: malformed hyperlink target. [docutils]
```
and the build fails. Hopefully RTD sees the same...

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -19,9 +19,7 @@ import sys
 # import eelbrain.plot._brain_object  # make sure that Brain is available
 import eelbrain
 from sphinx_gallery.sorting import _SortKey
-import sphinx.util.logging
 
-logger = sphinx.util.logging.getLogger("eelbrain")
 need_path = str(Path(__file__).parent)
 if need_path not in sys.path:
     sys.path.append(need_path)  # for conf.NameOrder and conf.use_pyplot
@@ -68,6 +66,7 @@ bibtex_bibfiles = ['publications.bib']
 qualname_overrides = {
     "eelbrain._data_obj.NDVar": "eelbrain.NDVar",
 }
+
 
 ################################################################################
 # Sphinx-Gallery
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -14,8 +14,6 @@ import os
 from pathlib import Path
 import sys
 
-# TODO: This just warns, so doesn't actually make sure Brain is available:
-# import eelbrain.plot._brain_object  # make sure that Brain is available
 import eelbrain
 from sphinx_gallery.sorting import _SortKey
 import eelbrain.datasets._alice
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -78,19 +78,19 @@ def use_pyplot(gallery_conf, fname):
 
 
 example_order = {
-    'datasets': [
+    '../examples/datasets': [
         'intro.py',
         'dataset-basics.py',
         'align.py',
         'ndvar-creating.py',
         'group-level-analysis.py',
     ],
-    'statistics': [
+    '../examples/statistics': [
         'models.py',
         'RMANOVA.py',
         'ANCOVA.py',
     ],
-    'mass-univariate-statistics': [
+    '../examples/mass-univariate-statistics': [
         'permutation-statistics.py',
         'sensor-ttest.py',
         'sensor-anova.py',
@@ -99,14 +99,14 @@ example_order = {
         'compare-topographies.py',
         'volume-source-space.py',
     ],
-    'plots': [
+    '../examples/plots': [
         'boxplot.py',
         'utsstat.py',
         'colors.py',
         'colormaps.py',
         'customizing.py',
     ],
-    'temporal-response-functions': [
+    '../examples/temporal-response-functions': [
         'trf_intro.py',
         'alice-trf.py',
         'mtrf.py',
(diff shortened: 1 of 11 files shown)


doc/conf.py
```diff
@@ -13,11 +13,16 @@
 from itertools import chain
 import os
 from pathlib import Path
+import sys
 
-import eelbrain.plot._brain_object  # make sure that Brain is available
+# TODO: This just warns, so doesn't actually make sure Brain is available:
+# import eelbrain.plot._brain_object  # make sure that Brain is available
```
larsoner: I can either just get rid of this line, or adapt the env to actually install PySurfer...

christianbrodbeck: IIRC this was about triggering a lazy import, not about raising an error

larsoner: What's the goal? Currently all it does is cause a warning to be emitted that PySurfer can't be used. And this warning doesn't seem to be relevant for actually building the docs (it otherwise succeeds)...

It seems to me that either 1) if PySurfer needs to be installed, it should raise an error, or 2) if PySurfer doesn't need to be installed, the line should be removed. But maybe I misunderstand the purpose / utility here?

christianbrodbeck: My memory is that `sphinx` did not find `_brain_object.Brain` unless it had been imported before. We don't need PySurfer to create auto docs on `Brain`. This must be a while ago so maybe `sphinx` now does not need that anymore?

larsoner: Ahh okay... it might still need it. Someday someone could check... but for now I'll just catch the warning and update the comment!

larsoner: ... actually I guess I can just check the build once it succeeds, since I've already commented it out. I'll wait and check!

larsoner:
Okay looking at [PR](https://eelbrain--177.org.readthedocs.build/en/177/reference.html#module-plot.brain) vs [stable](https://eelbrain.readthedocs.io/en/stable/reference.html#module-plot.brain) neither one actually has `Brain` in the autosummaries, but both have the actual docs:

- https://eelbrain--177.org.readthedocs.build/en/177/generated/eelbrain.plot._brain_object.Brain.html#eelbrain.plot._brain_object.Brain
- https://eelbrain.readthedocs.io/en/stable/generated/eelbrain.plot._brain_object.Brain.html#eelbrain.plot._brain_object.Brain

So I guess it's okay to remove the line?

doc/reference.rst
```diff
@@ -528,7 +528,8 @@ More specific control over the plots can be achieved through the
    ~plot._brain_object.Brain
    plot.brain.SequencePlotter
    plot.GlassBrain
-   plot.GlassBrain.butterfly
+
+See also :meth:`plot.GlassBrain.butterfly`.
```
christianbrodbeck:
```suggestion
See also :meth:`plot.GlassBrain.butterfly` for a butterfly-plot with a time-linked :class:`~plot.GlassBrain` plot.
```

eelbrain/fmtxt.py
```diff
@@ -1969,6 +1969,22 @@ def save_image(self, dst):
         with open(dst, 'wb') as fid:
             fid.write(buf)
 
+    # Add this to work around a rendering issue
+    def seek(self, pos: int, whence: int = 0, /) -> int:
```
christianbrodbeck, line 1973: Isn't this already the default? https://docs.python.org/3/library/io.html#io.IOBase

christianbrodbeck:
@codex review

eelbrain/fmtxt.py
larsoner: Yes but its docstring has a rendering issue that Sphinx complains about. Really this class probably shouldn't have its inherited methods documented at all, but setting that up with sphinx and templates etc. was going to be a pain

larsoner: ... so this is just a wrapper with a `__doc__` that Sphinx is happy with

larsoner:
Suggestion committed!

eelbrain/fmtxt.py
```diff
@@ -1969,6 +1969,22 @@ def save_image(self, dst):
         with open(dst, 'wb') as fid:
             fid.write(buf)
 
+    # Add this to work around a rendering issue
```
christianbrodbeck:
```suggestion
    # Add this to work around a sphinx rendering issue with the inherited docstring
```

christianbrodbeck (dismissed):

christianbrodbeck (approved):

christianbrodbeck:
Thanks @larsoner !

PR #178, opened 2026-05-15
Fix Alice dataset download (DRUM WAF + URL updates)

christianbrodbeck:
## Summary

- Updates the `stimuli` and `eeg.1` bitstream download URLs to their new DRUM locations
- Adds a browser `User-Agent` header to the `pooch.HTTPDownloader` so the AWS WAF on drum.lib.umd.edu allows the download through (the default Python user-agent is blocked with 403)

## Test plan

- [ ] Run `eelbrain.datasets._alice.get_alice_path()` and confirm both zip files download and extract successfully

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/eelbrain/datasets/_alice.py
+++ b/eelbrain/datasets/_alice.py
@@ -12,14 +12,18 @@ def get_alice_path(
     path = Path(path).expanduser().resolve()
     path.mkdir(exist_ok=True, parents=True)
 
-    baseurl = 'https://drum.lib.umd.edu/bitstream/handle/1903/27591/'
     registry = {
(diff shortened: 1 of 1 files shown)


larsoner:
Looks like this duplicates effort from https://github.com/Eelbrain/Eelbrain/pull/133/changes#diff-584f499262cfa24339ea7406a1f1c287784605287d451108aa1d8db51ea5532eR27 :disappointed:

christianbrodbeck:
@larsoner sorry I missed that, for some reason I thought the doc build was still failing there

PR #193, opened 2026-06-30
Automatically set __version__ for release

christianbrodbeck:
This is meant to automatically set `__version__` from the release tag. @larsoner do you know of any reason this would not work?

### DIFF
--- a/.github/scripts/set_release_version.py
+++ b/.github/scripts/set_release_version.py
@@ -1,0 +1,29 @@
+"""Set eelbrain.__version__ from the current GitHub release tag."""
+import os
+import re
+from pathlib import Path
+
+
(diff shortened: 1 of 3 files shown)


larsoner:
I wouldn't try to do this stuff manually... `setuptools_scm` and `hatch-vcs` etc. have been used by many projects for a long time and have many unexected, annoying kinks worked out. I'm happy to open a PR to switch to that versioning scheme if you want. We used `setuptools_scm` for a year or two for MNE-Python and currently use `hatch-vcs`, but I still work with several projects that use `setuptools_scm`. Given you're using a `setuptools` in pyproject.toml I'd recommend `setuptools_scm` here

christianbrodbeck:
@larsoner If it's straight forward for you that would be great, thanks!!

PR #207, opened 2026-07-16
Add notch filter

tanya-jp:


christianbrodbeck:
Set to draft until CI pass

PR #210, opened 2026-07-20
FIX use a stable, content-derived code for Factor-valued event triggers

### LINKED ISSUE #209: `load_epoch()` fails when `value` column in BIDS sidecar ``*_events.tsv`` file contains string.
`EventsInput` expect  the ``value`` column  to contain integer trigger code, while BIDS specification does not apply any such restriction:

Image
As a result `load_epoch()` fails when string values are used.

Example:
Image
...
Image

proloyd:
RecordingEpochsDerivative.build() needs a numeric trigger for MNE's events array, but ds['value'] can legally become a Factor (e.g. via a user's label_events() hook or a LabelVar). The previous fix used the Factor's own internal codes (Var(_trigger.x)), which depend on which values happen to be observed in a given recording, so the same label could get different codes across runs. _factor_trigger_to_var() instead derives each code from a CRC32 hash of the label text itself, so the same label always maps to the same code with no shared state needed across recordings.

Addresses [#209](https://github.com/Eelbrain/Eelbrain/issues/209).


christianbrodbeck:
These `value` values are added to `mne.Epochs`, but are not used by the pipeline any further as they're redundant with the events in the enclosing `Dataset`. What about just dropping string labels when constructing epochs (`eelbrain._io.fiff._mne_events` already has a fallback for using `numpy.ones`)? Alternatively, if it would be useful to have the events also stored on the `Epoch`, then you could map them to their original string values using the `event_id` argument to `mne.Epochs`.

proloyd:
@christianbrodbeck yes, I noticed that pipeline is not using those event codes. I added it since I may use the `mne.Epochs`. Will add the `event_id` as well to preserve the mapping then.

eelbrain/_experiment/epochs/nodes.py
```diff
@@ -209,11 +209,13 @@ def build(self, ctx: Request):
         tmin, tmax, tstop, decim, variable_tmax = epoch._extraction_parameters(ds, ctx.options)
         # Baseline correction is deferred to a view operation and must not enter the cache,
         # except for post_baseline_trigger_shift epochs where it has to precede the shift.
+        # A Factor-valued 'value' (e.g. from label_events) is converted to a
+        # stable numeric trigger + matching event_id automatically.
         if variable_tmax:
-            epochs_list = load.mne.variable_length_mne_epochs(ds, tmin, tmax, None, allow_truncation=True, decim=decim, reject_by_annotation=False, i_start='sample', trigger='value')
```
christianbrodbeck, line 213: These changes should not be necessary anymore, right?

eelbrain/_io/tests/test_mne.py
```diff
@@ -117,6 +128,170 @@ def test_epochs_ndvar_ignores_bad_channels_for_auto_adjacency():
     assert tuple(ndvar.sensor.names) == ('Fz', 'Cz', 'Pz')
 
 
+def test_mne_epochs_event_id():
```
christianbrodbeck, line 131: Cold we combine these tests into a single test so that the data has to be generated only once? (or use a `@pytest.fixture`)

eelbrain/_io/fiff.py
```diff
@@ -587,6 +588,57 @@ def add_mne_epochs(ds, tmin=-0.1, tmax=None, baseline=None, target='epochs', **k
...
+    label always maps to the same code across different runs/recordings
+    without requiring a shared/external registry.
+
+    Also returns the label -> code mapping, for use as the ``event_id`` of
+    the resulting :class:`mne.Epochs`, so the original string labels remain
+    available to code that works with the raw ``mne.Epochs`` object directly.
+    """
+    code_of = {}
+    x = np.empty(len(factor), dtype=np.int64)
+    for i, label in enumerate(factor):
```
christianbrodbeck: Looping through `factor.cells` should be more efficient than looping through all values. Once you built the dictionary, you can then use `Factor.as_var`

eelbrain/_io/fiff.py
```diff
@@ -996,7 +1079,9 @@ def variable_length_mne_epochs(
             else:
                 missing = (i_max - raw.last_samp) / raw.info['sfreq']
                 raise ValueError(f"{tmax[i]=} is outside of data range by {missing:g} s")
-        epochs_i = mne.Epochs(raw, events_array[i:i + 1], None, tmin_i, tmax_i, baseline, picks, preload=True, decim=decim, **kwargs)
+        code_i = events_array[i, 2]
+        event_id_i = {label_of_code[code_i]: code_i} if code_i in label_of_code else None
```
christianbrodbeck:
```suggestion
        if code_i in label_of_code:
            event_id_i = {label_of_code[code_i]: code_i}
        else:
            event_id_i = None
```
I find these one-line if/else statements hard to read unless they're very short

christianbrodbeck (changes requested):
Thanks @proloyd ! Just some minor requests/suggestions

eelbrain/_io/fiff.py
proloyd: Great suggestion!

eelbrain/_experiment/epochs/nodes.py
proloyd: We still have to say trigger='value', no?  The created `ds` does not have any column called 'trigger' [the default value for trigger kwarg].

eelbrain/_io/tests/test_mne.py
proloyd: Taken care of now.

proloyd:
@christianbrodbeck, ready for another round of review.

christianbrodbeck:
@codex review

eelbrain/_io/fiff.py
```diff
@@ -587,6 +588,51 @@ def add_mne_epochs(ds, tmin=-0.1, tmax=None, baseline=None, target='epochs', **k
...
+    Each label's code is derived from the label's own text (CRC32), not from
+    which other values happen to co-occur in a given recording, so the same
+    label always maps to the same code across different runs/recordings
+    without requiring a shared/external registry.
+
+    Also returns the label -> code mapping, for use as the ``event_id`` of
+    the resulting :class:`mne.Epochs`, so the original string labels remain
+    available to code that works with the raw ``mne.Epochs`` object directly.
+    """
+    code_of = {label: zlib.crc32(label.encode()) & 0x7FFFFFFF for label in factor.cells}  # mask to a non-negative value that fits in int32
```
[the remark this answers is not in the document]
proloyd: Highly unlikely to happen.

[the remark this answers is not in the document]
christianbrodbeck: What about Murphy's Law :)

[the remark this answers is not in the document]
proloyd: Only an AI would use such a cryptic label. :/. And in that case, a human will have to clean up the mess.

The best I can do it to create a safe guard.

eelbrain/_io/fiff.py
```diff
@@ -587,6 +588,51 @@ def add_mne_epochs(ds, tmin=-0.1, tmax=None, baseline=None, target='epochs', **k
...
+    return factor.as_var(code_of), code_of
+
+
+def _resolve_trigger(ds, trigger, event_id):
```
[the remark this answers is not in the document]
proloyd: Addressed.

proloyd:
@christianbrodbeck address two of the points (P2 & P3). Left P1, since `ds` already keeps track of the events.

eelbrain/_io/fiff.py
```diff
@@ -587,6 +588,91 @@ def add_mne_epochs(ds, tmin=-0.1, tmax=None, baseline=None, target='epochs', **k
...
+    available to code that works with the raw ``mne.Epochs`` object directly.
+
+    Raises
+    ------
+    RuntimeError
+        If two distinct labels hash to the same code (astronomically
+        unlikely for realistic label counts, but would otherwise silently
+        merge two different conditions into one ``event_id``).
+    """
+    code_of = {label: zlib.crc32(label.encode()) & 0x7FFFFFFF for label in factor.cells}  # mask to a non-negative value that fits in int32
```
christianbrodbeck, line 613: **Cross-recording CRC32 collision is not detected.** `_factor_trigger_to_var`'s collision guard (raising `RuntimeError`) only inspects `factor.cells` for the single `ds`/recording passed to one call. It never checks against codes derived in *other* recordings/runs, or in a `SuperEpoch`'s other sub-epochs, whose `mne.Epochs` objects later get merged via `combine()` -> `mne.concatenate_epochs` (see `RecordingEpochsDerivative.build()` / `EpochsDerivative.build()` in `eelbrain/_experiment/epochs/nodes.py`).

Concretely: if recording A only has label `'target'` and recording B only has label `'foilXYZ'`, and their CRC32 hashes collide, each `_factor_trigger_to_var` call succeeds individually (no collision *within* either recording). When the resulting per-recording `mne.Epochs` are concatenated, `mne.concatenate_epochs`'s merge (`event_id.update(...)`) only raises when the **same key** maps to **different codes** across pieces — it does not catch **different keys mapping to the same code**. The combined `epochs.event_id` silently ends up with two different labels sharing one trigger code, so `epochs['target']`/`epochs['foilXYZ']` would select the same (mixed) set of events. This is exactly the kind of silent merge the new collision guard was added to prevent, but it doesn't cover the cross-recording case — which is the PR's own stated primary use case ("the same label always maps to the same code... across different runs/recordings").

eelbrain/_io/fiff.py
```diff
@@ -978,7 +1091,11 @@ def variable_length_mne_epochs(
         tmax = np.repeat(tmax, n)
     if picks is None and raw.info['bads']:
         picks = mne.pick_types(raw.info, meg=True, eeg=True, eog=True, ref_meg=False, exclude=[])
+    trigger, event_id = _resolve_trigger(events, trigger, event_id)
     events_array = _mne_events(events, i_start=i_start, trigger=trigger)
+    # mne.Epochs requires every event_id value to have a matching event, so
+    # each single-epoch call below only gets the one entry for its own code
+    label_of_code = {code: label for label, code in (event_id or {}).items()}
```
christianbrodbeck, line 1098: **Silent mislabeling when an explicit `event_id` has duplicate code values.** `label_of_code = {code: label for label, code in (event_id or {}).items()}` inverts the dict; if the caller-supplied `event_id` maps two different labels to the same numeric code (e.g. `{'a': 1, 'b': 1, 'c': 3}`), this keeps only the last-inserted label for that code. Down at line ~1119, `event_id_i = {label_of_code[code_i]: code_i}` then attaches that (possibly wrong) label to every epoch whose trigger code is 1 — including the ones that were actually labeled `'a'`.

This was empirically verified: for a trigger sequence `['a','b','c']` with codes `[1,1,3]`, `variable_length_mne_epochs` returns per-epoch `event_id` of `[{'b': 1}, {'b': 1}, {'c': 3}]` — the epoch whose real label is `'a'` is mislabeled `'b'` in its own `mne.Epochs.event_id`, with no warning. This also silently diverges from `mne_epochs` (the fixed-length/single-call path), which passes the caller's `event_id` dict through unchanged.

eelbrain/_io/fiff.py
```diff
@@ -587,6 +588,91 @@ def add_mne_epochs(ds, tmin=-0.1, tmax=None, baseline=None, target='epochs', **k
...
+        ``trigger`` or dropped (with a warning) for being incompatible with a
+        ``None`` trigger.
+    """
+    if isinstance(trigger, str):
+        trigger = ds[trigger]
+    if isinstance(trigger, Factor):
+        if event_id is None:
+            trigger, event_id = _factor_trigger_to_var(trigger)
+        else:
+            trigger = Var([event_id[label] for label in trigger])
```
christianbrodbeck, line 666: **Bare `KeyError` when an explicit `event_id` doesn't cover every label in the Factor.** `trigger = Var([event_id[label] for label in trigger])` does a strict dict lookup per event; if any label produced by the Factor-valued trigger (e.g. from `label_events`) is missing from a caller-supplied `event_id`, this raises an opaque `KeyError: '<label>'` deep inside epoch construction instead of a descriptive error. This is inconsistent with the CRC32-collision path a few lines above, which raises a clear `RuntimeError` naming the offending labels. Example: `mne_epochs(ds, trigger='value', event_id={'target': 1})` where `ds['value']` also contains the label `'distractor'`.

eelbrain/_io/fiff.py
```diff
@@ -587,6 +588,91 @@ def add_mne_epochs(ds, tmin=-0.1, tmax=None, baseline=None, target='epochs', **k
...
+    elif trigger is None and event_id is not None and set(event_id.values()) != {1}:
```
christianbrodbeck, line 667: **Empty `event_id={}` is spuriously treated as incompatible with `trigger=None`.** `set(event_id.values()) != {1}` evaluates to `True` when `event_id` is `{}` (since `set() != {1}`), so passing `trigger=None, event_id={}` triggers the "a None trigger assigns every event the code 1..." warning and discards `event_id`, even though an empty dict declares no conflicting codes at all. Minor, but a real logic gap in the condition (should probably be `event_id.values() - {1}` truthy, or an explicit `event_id` emptiness check).

eelbrain/_io/fiff.py
christianbrodbeck, line 613: **Cleanup:** the CRC32 code is masked to 31 bits (`& 0x7FFFFFFF`), discarding the top bit of the 32-bit hash and halving the collision-resistant space (2^31 vs 2^32) even though the full unsigned CRC range could be reinterpreted as a signed int32 instead. This isn't itself a correctness bug, but it needlessly increases the odds of hitting the collision case flagged above (both within one recording, which is now guarded, and across recordings, which isn't).

eelbrain/_io/fiff.py
christianbrodbeck, line 666: **Reuse:** `Var([event_id[label] for label in trigger])` re-implements `Factor.as_var()` / `_Effect.as_var()` (`eelbrain/_data_obj.py`), which already does `Var([labels[v] for v in self], name)`. Prefer `trigger.as_var(event_id)` — same behavior (including the `KeyError` on a missing label) but avoids a second, driftable copy of the conversion, and preserves the Var's `name` which the manual version drops.

eelbrain/_experiment/events.py
```diff
@@ -113,9 +113,9 @@ class EventsInput(Input[Dataset]):
 
     The file is expected to follow the BIDS specification with at least the
     columns ``onset`` (seconds), ``sample`` (integer sample index), and
-    ``value`` (integer trigger code).  Additional columns are passed through
-    to the returned :class:`~eelbrain.Dataset` so that they are available in
-    :meth:`~Pipeline.label_events`.
+    ``value`` (trigger code, interger or strings).  Additional columns are
```
christianbrodbeck, line 116: Typo: "interger" should be "integer" (and probably "integer or string" rather than "integer or strings").

eelbrain/_io/fiff.py
christianbrodbeck: FYI these comments are from Claude but it used my GH CLI to post them hence they appear under my name.

Not sure how to think about the possible hash collision -- it's very low probability but unless I'm misreading the cost could be very high (a spurious result that is not based on the events that the user intends to use).

Could we set the epoch's event_id at the Pipeline load point (after recordings have been merged)?

proloyd: It is very probability, and would happen with any such hashing. I don't think we need to worry about it too much.

eelbrain/_io/fiff.py
proloyd: `labels_of_code` now keeps all labels per code instead.

eelbrain/_io/fiff.py
proloyd: Will raise a descriptive `KeyError` for `event_id` missing a `Factor`-trigger label.

proloyd: Use `Factor.as_var()`.

eelbrain/_io/fiff.py
proloyd: Will treat an empty `event_id` like `None` in `_resolve_trigger`.

eelbrain/_experiment/events.py
proloyd: Taken care of.

eelbrain/_io/fiff.py
proloyd: I don't think we will have to go that way. the collision chances are very little.

eelbrain/_io/fiff.py
proloyd: Sorry my bad. MNE-Python 1.13 does not allow duplicate values in the event_id dictionary. So, we now raise a `ValueError` naming the colliding labels.

proloyd:
@christianbrodbeck ready for your review.

eelbrain/_io/fiff.py
proloyd: Also, that claim is simply wrong for the MNE-Python 1.13.2. `mne.concatenate_epochs` now also raises when different keys maps to the same code.

PR #218, opened 2026-08-21
Update fiff.py

nigelflower:
Subtracted raw.first_samp from event onsets


christianbrodbeck:
Setting to draft until we figure out the underlying issue

PR #220, opened 2026-08-24
Pipeline: add by-predictor TRF lags

christianbrodbeck:
@nigelflower this should work without invalidating an existing cache

- [x] What happens with `@` comparison syntax?


christianbrodbeck:
@huzafahyde this now implements the variables time lags you're working with

christianbrodbeck:
Realizing that it might be non-obvious where to find how to use this I started updating the docs in #221. Once the are built, look for *Lag Windows* under *The Pipeline*/*TRF Analysis*

PR #222, opened 2026-08-25
Pipeline: use cHPI

christianbrodbeck:
@nigelflower @jpereznavarro

This should enable cHPI in the Pipeline. Requires current `main` branch of mne to work correctly.


christianbrodbeck:
@nigelflower @jpereznavarro, this is ready for testing.
@jpereznavarro, if you computed any results with an intermediate branch state it may be safer to delete the cache first.

christianbrodbeck:
@nigelflower @jpereznavarro, this now addresses the mne warning about small condition number. For subjects with acceptable condition number, nothing should change, but for subjects with small condition number the covariance will be regularized automatically. Note that I set the threshold a bit higher than mne. If you don't want subjects to be affected by this that did not previously issue the warning then you could override the `emptyroom` covariance (see https://eelbrain--222.org.readthedocs.build/en/222/pipeline/source.html#eelbrain.pipeline.Pipeline.noise_covariance)

PR #237, opened 2026-09-23
Pipeline: exclude runs in the concatenated epochs according to `ignore_entities`

### LINKED ISSUE #236: [Bug]: 'ignore_runs' of `ignore_entities` are not getting honored.
### Description

`find_matching_paths` does accept a `runs=` parameter, but eelbrain's `Pipeline.__init__` doesn't pass it — that's the bug.

The fix is to pass `runs=self._runs or None` to `find_matching_paths`.

### Minimal reproducible example

```
class ImageNet(Pipeline):
    accession_no = 'ds005810'

    preload = True

    ignore_entities = {
        'ignore_subjects': ('emptyroom'),
        'ignore_sessions': ('ImageNet01', 'ImageNet02', 'ImageNet04', 'MRI'),
        'ignore_runs': ('06', '07', '08', '09', '10'),
    }

    raw = {
        'raw': RawSource(adjaceny='ctf275'),
    }

    epochs = {
        'cont': ContinuousEpoch(task='ImageNet', samplingrate=200),
    }

e = ImageNet(rootdir)
ds = e.load_epochs(session='ImageNet03', epoch='cont')
ds.head(-1)
```

### Expected behavior

Should have ignored the ignored runs, i.e.,

Image

### Actual behavior

reads all the runs.

Image

### Environment details

MacOS

### Additional context

_No response_

proloyd:
Resolves [Bug]: 'ignore_runs' of ignore_entities are not getting honored. #236


christianbrodbeck:
@codex review

christianbrodbeck (commented):
Reviewed ee18c62761d3fe3dd96290530a2f8009e6d44aa4. Found one regression affecting datasets that mix numbered runs with recordings without a run entity (inline comment).

Validation: all four targeted tests passed (`test_acquisition.py` and `test_sample_experiment.py::test_epoch_run`). Also reproduced the mixed-run failure with synthetic FIF recordings: ICA input loading raises `FileMissingError` on the PR, while restoring the original discovery call loads the 200-sample recording successfully. Tests used the existing eeldev environment and its prebuilt Eelbrain extensions; the full suite was not run.

eelbrain/_experiment/pipeline.py
```diff
@@ -302,7 +302,7 @@ def __init__(
         acquisitions = tuple(get_entity_vals(root, 'acquisition', **ignore_entities))
         matching_paths = tuple(
             path
-            for path in find_matching_paths(root, subjects=self._subjects, sessions=self._sessions, tasks=self._tasks, datatypes=datatype, suffixes=datatype, extensions=extensions, ignore_nosub=True)
+            for path in find_matching_paths(root, subjects=self._subjects, sessions=self._sessions, tasks=self._tasks, runs=self._runs or None, datatypes=datatype, suffixes=datatype, extensions=extensions, ignore_nosub=True)
```
christianbrodbeck: [P2] Preserve recordings without a run entity when filtering runs

`get_entity_vals(..., 'run')` only returns explicit run labels, so passing that tuple to `find_matching_paths` also excludes every recording without a `run-*` entity whenever any numbered run exists elsewhere in the dataset—even with the default `ignore_entities = {}`. For example, with `sub-01_task-test_run-1_meg.fif`, `sub-01_task-test_run-2_meg.fif`, and `sub-02_task-test_meg.fif`, subject 02 disappears from `_recordings`. With `RawICA('raw', ...)`, its `_source_states()` becomes empty and `load_concatenated_source_raw()` raises `FileMissingError: No source recordings found to estimate ICA`; restoring the original discovery call successfully loads that same recording. Please preserve paths with `path.run is None` while excluding ignored numbered runs (analogous to the acquisition predicate below), and add a mixed numbered/unnumbered recording regression test.

